PR #16, opened 2017-07-20
Better error unknown

kevmoo:


lib/src/json_serializable_generator.dart
```diff
@@ -58,6 +58,17 @@ class JsonSerializableGenerator
     // TODO: support overriding the field set with an annotation option
     var fieldsList = classElement.fields.where((e) => !e.isStatic).toList();
 
+    var undefinedFields =
+        fieldsList.where((fe) => fe.type.isUndefined).toList();
```
natebosch: why do you need toList?

lib/src/json_serializable_generator.dart
```diff
@@ -151,6 +162,19 @@ class JsonSerializableGenerator
       fieldsToSet.remove(arg.name);
     }
 
+    var undefinedArgs = [ctorArguments, ctorNamedArguments]
+        .expand((l) => l)
```
natebosch: we need Iterable.followedBy :)

lib/src/json_serializable_generator.dart
```diff
@@ -151,6 +162,19 @@ class JsonSerializableGenerator
...
+        .where((pe) => pe.type.isUndefined)
+        .toList();
```
natebosch: why toList()?

lib/src/json_serializable_generator.dart
```diff
@@ -296,15 +320,17 @@ class JsonSerializableGenerator
       {ParameterElement ctorParam}) {
     name = _fieldToJsonMapKey(name, field);
     var result = "json['$name']";
-    return _writeAccessToJsonValue(result, field.type, ctorParam: ctorParam);
+    DartType targetType;
```
natebosch: [optional] `var targetType = ctorParam?.type ?? field.type;` or `var targetType = (ctorParam ?? field).type;`

lib/src/json_serializable_generator.dart
kevmoo: Indeed!

lib/src/json_serializable_generator.dart
kevmoo: I hate reiterating iterables – even trivially. Can have unintended perf and behavior affects.

So I just never do it.

lib/src/json_serializable_generator.dart
kevmoo: ditto

lib/src/json_serializable_generator.dart
kevmoo: nice

PR #26, opened 2017-07-25
Option to omit empty fields

kevmoo:


lib/src/json_serializable.dart
```diff
@@ -6,9 +6,17 @@ class JsonSerializable {
   final bool createFactory;
   final bool createToJson;
 
-  const JsonSerializable({bool createFactory: true, bool createToJson: true})
-      : this.createFactory = createFactory,
-        this.createToJson = createToJson;
+  /// [true] if the generator should include the this field in the serialized
```
natebosch: [nit] 'Whether the generator should...'

lib/src/json_serializable.dart
```diff
@@ -6,9 +6,17 @@ class JsonSerializable {
...
+  /// output, even if the value is `null`.
+  final bool includeIfNull;
+
+  const JsonSerializable(
+      {bool createFactory: true,
+      bool createToJson: true,
+      bool includeIfNull: true})
```
natebosch: why is true the default? Backwards compatibility?

kevmoo: Yup. It can also lead to some weirdness – like having a completely empty Map returned from toJson – better to make it opt-in.

Also the code generated is a lot uglier.

lib/src/json_serializable.dart
kevmoo: ack

PR #29, opened 2017-07-27
Add support for enums

kevmoo:
Fixes https://github.com/dart-lang/json_serializable/issues/12

### DIFF
--- a/lib/src/type_helpers/type_helper_utils.dart
+++ b/lib/src/type_helpers/type_helper_utils.dart
@@ -1,0 +1,5 @@
+String commonNullPrefix(
+        bool nullable, String expression, String unsafeExpression) =>
+    nullable
+        ? "$expression == null ? null : $unsafeExpression"
+        : unsafeExpression;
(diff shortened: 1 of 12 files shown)


lib/src/type_helpers/type_helper_utils.dart
```diff
@@ -0,0 +1,7 @@
+String commonNullPrefix(
```
natebosch: [optional]

```dart
String commonNullPrefix(
        bool nullable, String expression, String unsafeExpression) =>
    nullable
        ? "$expression == null ? null : $unsafeExpression"
        : unsafeExpression;
```

test/test_files/json_test_example.dart
```diff
@@ -29,16 +33,21 @@ class Person extends Object with _$PersonSerializerMixin {
       dateOfBirth == other.dateOfBirth;
 }
 
+enum Category { top, bottom, strage, charmed, up, down }
```
natebosch: 👍 (except the typo in 'strange')

lib/src/type_helpers/type_helper_utils.dart
kevmoo: You're a machine.

PR #30, opened 2017-08-01
Fix Dart formatting in JSON strings, standardized strings, etc

kevmoo:


lib/src/json_literal_generator.dart
```diff
@@ -28,14 +28,65 @@ class JsonLiteralGenerator extends GeneratorForAnnotation<JsonLiteral> {
...
+    var thing = _jsonLiteralAsDart(content, asConst).toString();
+    var marked = asConst ? 'const' : 'final';
 
     return '$marked _\$${element.displayName}JsonLiteral = $thing;';
   }
 }
 
-bool _isConstType(value) {
-  return value == null || value is String || value is num || value is bool;
+String _jsonLiteralAsDart(dynamic value, bool asConst) {
```
natebosch: maybe break out some methods?

This is really hard to follow. Also a Doc comment would be nice

kevmoo: PTAL

lib/src/json_literal_generator.dart
```diff
@@ -28,14 +28,88 @@ class JsonLiteralGenerator extends GeneratorForAnnotation<JsonLiteral> {
...
 
-bool _isConstType(value) {
-  return value == null || value is String || value is num || value is bool;
+/// Returns a [String] representing a valid Dart literal for [value].
+///
+/// If [asConst] is `true`, the returned [String] is encoded as a `const`
+/// literal.
+String _jsonLiteralAsDart(dynamic value, bool asConst) {
+  if (value == null) {
+    return 'null';
```
natebosch: It might read easier if we don't use the `else` after a return. Also could optionally inline the return

```dart
if (value == null) return 'null';
if (value is String) return _jsonStringAsDart(value);
if (value is bool || value is num) {
  return value.toString();
}
if (value is List) return _jsonListAsDart(value, asConst);
if (value is Map) return _jsonMapAsDart(value, asConst);
throw ...
```

kevmoo: I generally like closing braces...but I see your point...

PR #31, opened 2017-08-15
WIP: Trying to support the latest version of pkg/build

kevmoo:
See failures – your bug or mine?

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -7,12 +7,12 @@ environment:
   sdk: '>=1.24.0 <2.0.0-dev.infinity'
 dependencies:
   analyzer: '>=0.29.10 <0.31.0'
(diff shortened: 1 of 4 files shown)


natebosch:
It's entirely possible the new build_runner slowed things down though I wouldn't expect by much - is it deadlocking altogether? If so it's probably a build_runner bug

kevmoo:
I just pushed a patch to that PR that sets the exit code

kevmoo:
@natebosch see here https://travis-ci.org/dart-lang/json_serializable/jobs/264799208#L528

natebosch:
Definitely a bug in build_runner.

natebosch:
Ok, pub upgrade and this should work now.

PR #33, opened 2017-08-21
Improve toJson generation with some no-write null fields

kevmoo:


example/example.g.dart
```diff
@@ -26,20 +26,22 @@ abstract class _$PersonSerializerMixin {
...
 
-    $map['firstName'] = firstName;
-    $writeNotNull('middleName', middleName);
-    $map['lastName'] = lastName;
-    $map['date-of-birth'] = dateOfBirth.toIso8601String();
-    $map['last-order'] = lastOrder?.toIso8601String();
-    $map['orders'] = orders;
-    return $map;
+    writeNotNull('middleName', middleName);
+    val['lastName'] = lastName;
```
natebosch: can/should we get these other non-null fields into the initial map?

example/example.g.dart
```diff
@@ -26,20 +26,22 @@ abstract class _$PersonSerializerMixin {
   DateTime get lastOrder;
   List<Order> get orders;
   Map<String, dynamic> toJson() {
-    var $map = <String, dynamic>{};
-    void $writeNotNull(String key, dynamic value) {
+    var val = <String, dynamic>{
+      'firstName': firstName,
+    };
+
+    void writeNotNull(String key, dynamic value) {
```
natebosch: taking a look at this - I might skip the extra method and inline the conditionals through the rest. It's noisy but shouldn't matter since it's generated code

example/example.g.dart
kevmoo: Sadly, no. We want to keep the fields in consistent order.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.2.3
+
+* Write out `toJson` methods more efficiently the first fields written are not
```
kevmoo: fixed

example/example.g.dart
kevmoo: Hrm. Maybe. Will look at in another.

PR #36, opened 2017-08-21
WIP: remove toJson helper – not sure I like this...

kevmoo:
I think the runtime cost of the inline method must be pretty small – likely inlined in dart2js.


kevmoo:
Eh...I'm not moved enough to want to land it...

PR #39, opened 2017-09-25
Fixes to lints and fuzzy arrows

kevmoo:


lib/src/json_literal_generator.dart
```diff
@@ -54,21 +54,21 @@ String _jsonLiteralAsDart(dynamic value, bool asConst) {
     return '${asConst ? 'const' : ''}[$listItems]';
   }
 
-  if (value is Map) return _jsonMapAsDart(value, asConst);
+  if (value is Map<String, dynamic>) return _jsonMapAsDart(value, asConst);
```
jakemac53: Are these types of checks any more sane than they used to be (won't this return true for `Map<dynamic, dynamic>`)?

I guess you are just trying to avoid the explicit cast given the tighter type on `_jsonMapAsDart`, but this sends a bit of the wrong signal imo... I would rather see the cast when you pass the value in to `_jsonMapAsDart` as its a bit more obvious that you are explicitly casting it and it could technically fail.

PR #49, opened 2017-10-03
Centralize error handling for calls to _serialize

kevmoo:


kevmoo:
Centralize try-catch blocks and error strings – it gets silly w/ the custom toJson map work without this

PR #59, opened 2017-10-09
source_gen custom header – for copyright

kevmoo:


json_serializable/tool/build_actions.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
natebosch: I think we don't generally bother updating years when we touch a file?

kevmoo:
Felt weird to put 2017 in the text block and not update the file. 🤷‍♂️

PR #70, opened 2017-11-13
Pass field metadata to type helpers.

caleo-git:
We're struggling from the lack of control over serialization of fields. We need a way to get field annotations in custom type helpers to improve logic of serialization.


kevmoo:
@caleo-git want to try again? Just landed the required PR

PR #72, opened 2017-11-14
allow optional parameters for types using fromJson()

johnpryan:
I'm not sure this should really be allowed or should show a warning, but there are cases where `fromJson` constructors may implement optional parameters:

for example:

```
class Foo {
  String name;
  int value;
  Foo.fromJson(json, {initValue: false}) {
    name = json['name'];
    value = json['value'];
    if (initValue) {
      value = value ?? 0;
    }
  }
}
```

```
@JsonSerializable()
class Person extends Object with _$PersonSerializerMixin {
  final Foo foo;

  Person(this.foo);

  factory Person.fromJson(Map<String, dynamic> json) => _$PersonFromJson(json);
}

class Foo {
  String name;
  int value;
  Foo.fromJson(json, {initValue: false}) {
    name = json['name'];
    value = json['value'];
    if (initValue) {
      value = value ?? 0;
    }
  }
}
```

currently throws an exception because `Foo.fromJson` does not have a single parameter.


johnpryan:
Thanks for the quick response

1. done
2. done

json_serializable/test/json_serializable_test.dart
```diff
@@ -188,6 +188,14 @@ abstract class _$OrderSerializerMixin {
       });
     }
 
+
+    test('class with fromJson() constructor with optional parameters', () async {
+      var output = await runForElementNamed('FromJsonOptionalParameters');
+      print(output);
```
kevmoo: Remove this guy!

johnpryan: oh man... fixed.

I decided to squash the changes to hide my shame...

PR #80, opened 2017-12-13
Creating a stand-alone example

kevmoo:


kevmoo:
CC @sethladd

README.md
```diff
@@ -1,7 +1,7 @@
 # [![Build Status](https://travis-ci.org/dart-lang/json_serializable.svg?branch=master)](https://travis-ci.org/dart-lang/json_serializable)
 
-Provides `source_gen` `Generator`s which generate code to
-make it simple to serialize to and from JSON.
+Provides [source_gen] `Generator`s to create code for JSON serialization and
+deserialization.
```
sethladd: Can we clearly state something like "To use this package, you need to install source_gen, by going here." ?

README.md
```diff
@@ -15,5 +15,12 @@ The annotation package which has no dependencies.
 
 Import it into your pubspec `dependencies:` section.
 
+## [example]
+
+A simple example showing how to setup and use [json_serializable] and
```
sethladd: I'm confused, where is the example?

README.md
kevmoo: There's no install, per se. You just put the right things in your pubspec.

I will make sure to mention that.

README.md
kevmoo: See https://github.com/dart-lang/json_serializable/tree/example (although the link is broken because this isn't on master yet)

example/README.md
```diff
@@ -0,0 +1,23 @@
...
+  `JsonSerializable` and `JsonLiteral` – to enable code generation.
+
+  * Note: the annotations are defined in `package:json_annotation`.
+    This is the only package required in the `dependencies` section of your
+    `pubspec.yaml`.
+
+* `tool/`: Contains the code run during development to create and update
+  generated code.
+
+  * `build_actions.dart`: A convention when using `package:build` to
```
jakemac53: We actually should change this convention - having separate scripts means they have separate asset graphs which isn't what you want.

Instead you could add a flag to change whether it calls build/watch?

example/test/example_test.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: 2017

example/README.md
kevmoo: Then I have to add args parsing, etc to the example.

I'd rather have https://github.com/dart-lang/build/issues/718

example/test/example_test.dart
kevmoo: done!

kevmoo:
> I don't like having an example which isn't putting forward best practices - but we can improve it once we have dart-lang/build#718.

Let's make best practices easier first! 😄

PR #81, opened 2017-12-14
clarify what source_gen is

sethladd:


matanlurey:
Does this package "require" source_gen? It's an implementation detail.

sethladd:
That's what I'm confused about.

I land on this README and I don't know how to use this.

kevmoo:
There's more work to do here, but this commit doesn't get there.

Will take some time today to dig in...

PR #85, opened 2017-12-15
No source gen dep

kevmoo:
I will update the example once I publish the latest of each package


json_serializable/lib/src/json_part_builder.dart
```diff
@@ -0,0 +1,42 @@
...
+///   'my_package',
+///   inputs: const ['lib/*.dart']
+/// )
+/// ```
+///
+/// See the [example] to understand how to configure your project for
+/// `json_serializable`.
+///
+/// [example]: https://github.com/dart-lang/json_serializable/tree/master/example
+PartBuilder jsonPartBuilder({String header, bool useWrappers: false}) {
```
natebosch: why do you need to expose this as `PartBuilder`? Do you want to expose it instead as a `Builder`?

kevmoo: Yup!

PR #86, opened 2017-12-15
Prepare to release json_annotation

kevmoo:


kevmoo:
Then I can bump json_serializable and use the right dependency, etc etc...

PR #88, opened 2017-12-15
example: update to json_serializable 0.3.0

kevmoo:
Removed
  1 dev dependency
  12 lines from tool/build_actions.dart

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -3,12 +3,11 @@ dependencies to your `pubspec.yaml`.
 
 ```yaml
 dependencies:
-  json_annotation: ^0.2.1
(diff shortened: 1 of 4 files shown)


kevmoo:
CC @sethladd

sethladd:
Cool? Not sure how to evaluate this, but less dependencies seems like always a great idea!

kevmoo:
> Cool? Not sure how to evaluate this, but less dependencies seems like always a great idea!

Fewer things to get wrong, etc. Docs will help here, too.

Notice how `source_gen` disappeared? It's just an impl detail for `json_serializable`. Fewer things for the user to understand...

PR #90, opened 2017-12-22
Upgrade to build_runner 0.7

natebosch:
Note: Using git dependencies to send this early - will publish
build_runner before submitting.

- Update dependencies
- Add a `build.yaml` in `json_serializable` to expose a builder that is
  applied automatically to dependents of this package.
- Add a `builder.dart` to expose a `BuilderFactory`.
- Add a `build.yaml` in `example` to configure the header and keep the
  same behavior as the manual build script.
- Update the README to use the new approach.
- Keep manual build scripts for `json_serializable` since it does
  strange stuff. Update to the `applyToRoot` approach.


natebosch:
@kevmoo - build_runner latest is dev only SDK. Not sure how you feel about dropping travis testing for stable SDK...

kevmoo:
> @kevmoo - build_runner latest is dev only SDK. Not sure how you feel about dropping travis testing for stable SDK...

Well if it's not going to work, we should probably stop testing it 😄

json_serializable/pubspec.yaml
```diff
@@ -8,6 +8,7 @@ environment:
 dependencies:
```
kevmoo: @natebosch – should update the SDK dependency then, too? Right?

json_serializable/build.yaml
```diff
@@ -0,0 +1,8 @@
+builders:
```
kevmoo: We should add a `# some link here` comment to point source diggers to the right spot to understand how this is suppose to work?

json_serializable/tool/build_actions.dart
```diff
@@ -5,50 +5,46 @@
 import 'dart:async';
```
kevmoo: @natebosch – want to update to the new `run` method and go from 3 -> 1 file? Or in a follow-up?

example/build.yaml
```diff
@@ -0,0 +1,12 @@
+targets:
```
kevmoo: Ditto – add a `# FYI ready about build.yaml here...` line at the top...

json_serializable/pubspec.yaml
natebosch: IMO it shouldn't be a requirement. It's pub's job to take dependency SDK constraints into account when version solving.

If I'm only using stuff from `package:foo` version `1.0.0` but my dependency tightened the constraint to `1.1.0` I don't go hunting for that information and bump tighten my constraint too - that's what the version solve is for. From my perspective the SDK constraint is just like a package constraint in this way.

In practice I think it's fine to do it since we're not testing on older SDKs so we're not confident that we haven't suddenly stopped working for reasons around _this_ package.

example/build.yaml
natebosch: Done

json_serializable/build.yaml
natebosch: Done

json_serializable/pubspec.yaml
natebosch: Done

json_serializable/tool/build_actions.dart
natebosch: Done

PR #104, opened 2018-02-15
Cleanup up some lints when generating

kevmoo:
And other fixes

### DIFF
--- a/json_serializable/lib/src/type_helpers/enum_helper.dart
+++ b/json_serializable/lib/src/type_helpers/enum_helper.dart
@@ -29,7 +29,7 @@ class EnumHelper extends TypeHelper {
           nullable,
           expression,
           '$targetType.values.singleWhere('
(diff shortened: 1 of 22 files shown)


json_serializable/lib/src/type_helpers/enum_helper.dart
```diff
@@ -20,11 +26,18 @@ class EnumHelper extends TypeHelper {
   String deserialize(DartType targetType, String expression, bool nullable,
       DeserializeContext context) {
     if (targetType is InterfaceType && targetType.element.isEnum) {
+      String wrappedExpression;
```
natebosch: [optional] This type of flow is precisely what the `?:` operator is designed for - it makes it clear that the condition surrounds assignment rather than some other side effects.

`var wrappedExpressions = simpleExpressions.hasMatch(expression) ? expression : '{$expression}';`

kevmoo: ack

PR #107, opened 2018-02-21
Support inherited fields

kevmoo:


json_serializable/lib/src/json_serializable_generator.dart
```diff
@@ -558,3 +583,28 @@ InvalidGenerationSourceError _createInvalidGenerationError(
   return new InvalidGenerationSourceError(message,
       todo: 'Make sure all of the types are serializable.');
 }
+
+/// Returns a list of all instance, [FieldElement] items for [element] and
+/// super classes.
+List<FieldElement> _getFields(ClassElement element) {
```
natebosch: [optional] I have a personal vendetta against methods named `get*`. Maybe `findFields` or `allFields`?

json_serializable/lib/src/json_serializable_generator.dart
```diff
@@ -558,3 +583,28 @@ InvalidGenerationSourceError _createInvalidGenerationError(
...
+  // Get all of the fields that need to be assigned
+  // TODO: support overriding the field set with an annotation option
+  var fieldsList = element.fields.where((e) => !e.isStatic).toList();
+
+  var manager = new InheritanceManager(element.library);
+  var things = manager.getMembersInheritedFromClasses(element);
+
+  things.forEach((k, v) {
```
natebosch: you don't appear to use `k` at all and this loop has side effects. a for-in loop would be more idiomatic

```
for(final member in things.values) {
  assert(!member.isStatic)
  ...
}
```

json_serializable/lib/src/json_serializable_generator.dart
```diff
@@ -558,3 +583,28 @@ InvalidGenerationSourceError _createInvalidGenerationError(
...
+
+/// Returns a list of all instance, [FieldElement] items for [element] and
+/// super classes.
+List<FieldElement> _getFields(ClassElement element) {
+  // Get all of the fields that need to be assigned
+  // TODO: support overriding the field set with an annotation option
+  var fieldsList = element.fields.where((e) => !e.isStatic).toList();
+
+  var manager = new InheritanceManager(element.library);
+  var things = manager.getMembersInheritedFromClasses(element);
```
natebosch: things is not descriptive. `inheritedMembers` maybe, or skip the variable since it's not used more than once.

json_serializable/lib/src/json_serializable_generator.dart
```diff
@@ -558,3 +583,28 @@ InvalidGenerationSourceError _createInvalidGenerationError(
...
+
+  things.forEach((k, v) {
+    assert(!v.isStatic);
```
natebosch: are we asserting behavior of `getMembersInheritedFromClasses`? Do we need to?

json_serializable/lib/src/json_serializable_generator.dart
kevmoo: Changed to `_listFields`

json_serializable/lib/src/json_serializable_generator.dart
kevmoo: ack

json_serializable/lib/src/json_serializable_generator.dart
kevmoo: ack

json_serializable/lib/src/json_serializable_generator.dart
kevmoo: ack

PR #115, opened 2018-03-13
Made generator set field to the initial value if null

AlexanderJohr:
Hello Kevin Moore, from all the serialization libraries for dart out there, yours was always my favorite. But for me, it was lacking 2 major features: an ignore annotation and keep initial values if the deserialized field is null. I created two branches in this fork which include these features. With them, it's all I wanted from a serialization library for dart for a long time and I use it with joy already, but I want to share my joy with the community.

I want to add examples, tests, how to guides in the readme and so on. I also have to run the tests, which is difficult for me under windows, because the tools don't work so well and I'm lacking time to do it on Linux.

The reason for those pull requests is that I want to get in touch with you first and get feedback on how to change or rename certain things.

This pull request is for the feature that the generator sets the field to the initial value if the deserialized value is missing or null.

Without it all values are written to the fields even if the deserialized value is null or not found:
```dart
Column _$ColumnFromJson(Map<String, dynamic> json) =>
    new Column(containerId: json['containerId'] as String)
      ..settings =
          new ColumnSettings.fromJson(json['settings'] as Map<String, dynamic>);
```

With it, it checks if the deserialized value is null and if so it assigns the value that was already assigned to it.

```dart
Column _$ColumnFromJson(Map<String, dynamic> json) {
  Column column = new Column(containerId: json['containerId'] as String);
  column
    ..settings =
        new ColumnSettings.fromJson(json['settings'] as Map<String, dynamic>) ??
            column.settings;
  return column;
}
```

What I want to know is the following:

- This feature currently ignores the nullable annotation at all. This is because I don't know if there should be another name for the annotation as nullable already has the purpose to accept null values and let the user do the handling.
- same as includeIfNull = true, nullable = false causes _writeJsonValueNaive to be true and reduces the code size as null checks can be omitted. But in this case, nullable = false increases the code size of the FromJson factory constructor. `this` can't be used as in `..settings = valueFromJson ?? this.settings;` because `this.settings` doesn't exist in the mixing but in the class that uses the mixin. An extra object is needed for that and like that it's not a single statement anymore which adds the return statement and the curly braces.
- it would also be possible or even more performant to generate if(valueFromJson != null) blocks, because an extra assignment could be prevented like this but it would also increase the code size.

### DIFF
--- a/json_serializable/lib/src/json_serializable_generator.dart
+++ b/json_serializable/lib/src/json_serializable_generator.dart
@@ -383,9 +383,10 @@ void $toJsonMapHelperName(String key, dynamic value) {
     // Generate the static factory method
     //
     buffer.writeln();
+    String objectName = '${className[0].toLowerCase()}${className.substring(1)}';
     buffer
-        .writeln('$className ${prefix}FromJson(Map<String, dynamic> json) =>');
-    buffer.write('    new $className(');
+        .writeln('$className ${prefix}FromJson(Map<String, dynamic> json) {');
+    buffer.write('    $className $objectName = new $className(');
     buffer.writeAll(
         ctorArguments.map((paramElement) => _deserializeForField(
             fields[paramElement.name], classSupportNullable,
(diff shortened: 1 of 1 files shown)


kevmoo:
@AlexanderJohr – are you running with this one, too?

AlexanderJohr:
I guess I better close it for now and try it again, when my first pull request is finished.

PR #118, opened 2018-03-21
Introduced ignore attribute to JsonKey. When set to true the generato…

AlexanderJohr:
Hello Kevin Moore, from all the serialization libraries for dart out there, yours was always my favorite. But for me, it was a major feature: an ignore annotation. 

This pull request is for the ignore attribute for JsonKey. When set to true the generator excludes the field.

```dart
@JsonSerializable(includeIfNull: false, nullable: false)
class Column extends _$ColumnSerializerMixin {
// [...]
// ColumnSettings is serializable
ColumnSettings settings = new ColumnSettings();

// ColumnSettings is already serializable and includes witdh.
// Without 'ignore' the generator would generate the same field for ColumnSettings and Column 
@JsonKey(ignore: true)
num get width => settings.width;
set width (num value) =>settings.width = value;
// [...]
}
```

What is also does is:
- exclude private fields
- refactored _writeFactory to return the fields that it used to generate the factory instead of those which are not set by it.
- adds a test "works to ignore a field"
- adds a test 'includes final field in toJson when set in ctor' which ensures the final field 'a' that is set in the constructor is included in the factory and in toJson.
- adds a test 'excludes final field in toJson when not set in ctor' which ensures that the same field is excluded in factory and toJson if it isn't set in the constructor.

### DIFF
--- a/json_annotation/lib/src/json_serializable.dart
+++ b/json_annotation/lib/src/json_serializable.dart
@@ -65,10 +65,16 @@ class JsonKey {
   /// enclosing class.
   final bool includeIfNull;
 
+  /// `true` if the generator should ignore this field completely
+  ///
+  /// If `null` or `false`, the generator includes the field if it is not
+  /// sorted out by another condition.
+  final bool ignore;
+
   /// Creates a new [JsonKey].
   ///
   /// Only required when the default behavior is not desired.
-  const JsonKey({this.name, this.nullable, this.includeIfNull});
+  const JsonKey({this.name, this.nullable, this.includeIfNull, this.ignore});
 }
 
 // TODO(kevmoo): Add documentation
--- a/json_annotation/lib/src/json_serializable.dart
+++ b/json_annotation/lib/src/json_serializable.dart
@@ -65,10 +65,10 @@ class JsonKey {
   /// enclosing class.
   final bool includeIfNull;
 
-  /// `true` if the generator should ignore this field completely
+  /// `true` if the generator should ignore this field completely.
   ///
-  /// If `null` or `false`, the generator includes the field if it is not
-  /// sorted out by another condition.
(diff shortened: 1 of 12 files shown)


json_annotation/lib/src/json_serializable.dart
```diff
@@ -65,10 +65,16 @@ class JsonKey {
   /// enclosing class.
   final bool includeIfNull;
 
+  /// `true` if the generator should ignore this field completely
```
kevmoo: Add a `.` at the end of the sentence.

json_annotation/lib/src/json_serializable.dart
```diff
@@ -65,10 +65,16 @@ class JsonKey {
...
+  ///
+  /// If `null` or `false`, the generator includes the field if it is not
```
kevmoo:
```
If `null` (the default) or `false`, the field will be considered for serialization.
```

json_annotation/lib/src/json_serializable.dart
```diff
@@ -65,10 +65,16 @@ class JsonKey {
   /// enclosing class.
```
kevmoo: Also: bump the minor version of `json_annotation` in the pubspec to `0.2.3-dev`.

Add an entry to the changelog (under `0.2.3`) that includes this change.

AlexanderJohr:
What should I do with this hint from travis `hint • 'parameterKind' is deprecated and shouldn't be used at lib/src/json_serializable_generator.dart:355:17 • deprecated_member_use`

It's caused by an update of Dart 2 and I can't find info how to replace it.

kevmoo:
Rebase on this branch and you'll be good - https://github.com/dart-lang/json_serializable/tree/exclude_analyzer_alpha

AlexanderJohr:
Yes! Thank you for your reviews and your approval!

PR #120, opened 2018-03-30
latest pkg:source_gen, and a lot of deprecation cleanup

kevmoo:


kevmoo:
@natebosch – yes, this is a bear. Happy to split it up if it'll help...

json_serializable/build.yaml
```diff
@@ -2,8 +2,8 @@
 builders:
   json_serializable:
     target: ":json_serializable"
-    import: "package:json_serializable/builder.dart"
-    builder_factories: ["jsonSerializable"]
+    import: "package:json_serializable/json_serializable.dart"
+    builder_factories: ["createBuilderForJsonSerializable"]
```
natebosch: The changes in this file move away from the conventions we've been trying to establish. Perhaps I should document them somewhere...

Importing from `/builder.dart` is a signal that the code in here is meant more for the build system than for end users - it's good IMO to keep the builder separate from the annotations.

Having a noun name for the factory makes the build.yaml look a little nicer.

json_serializable/lib/src/builder.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch: 2018

json_serializable/lib/src/builder.dart
```diff
@@ -0,0 +1,30 @@
...
+import 'package:logging/logging.dart';
+
+import 'json_part_builder.dart';
+
+// TODO: until we can use `log` here - github.com/dart-lang/build/issues/1223
+final _logger = new Logger('json_serializable');
+
+/// Supports `package:build_runner` creation and configuration of `build_cli`.
+///
+/// Not meant to be invoked by hand-authored code.
```
natebosch: which is why I like to have it in a separate library 😁

json_serializable/build.yaml
kevmoo: Hrm...k

json_serializable/lib/src/builder.dart
kevmoo: Hrm...k

kevmoo:
@natebosch PTAL

PR #122, opened 2018-03-30
Refactor for reuse

kevmoo:
The only user-visible change here is fixing some error messages.

Otherwise, this just makes a few members public in a way that allows experimenting with reuse from other code.

Eventually, the code may end up in `source_gen` – but we'll take our time.

### DIFF
--- a/json_serializable/lib/src/constants.dart
+++ b/json_serializable/lib/src/constants.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
--- a/json_serializable/lib/src/constants.dart
+++ b/json_serializable/lib/src/constants.dart
@@ -1,0 +1,6 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+final toJsonMapVarName = 'val';
+final toJsonMapHelperName = 'writeNotNull';
--- a/json_serializable/lib/src/utils.dart
+++ b/json_serializable/lib/src/utils.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/analyzer.dart';
 import 'package:analyzer/dart/element/element.dart';
 
 // ignore: implementation_imports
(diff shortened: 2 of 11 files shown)


json_serializable/lib/src/constants.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
natebosch: 2018

json_serializable/lib/src/utils.dart
```diff
@@ -38,5 +51,182 @@ String friendlyNameForElement(Element element) {
...
+    }
+    return e.nameOffset;
+  }
+
+  return _offsetFor(a).compareTo(_offsetFor(b));
+}
+
+final _dartCoreObjectChecker = const TypeChecker.fromRuntime(Object);
+
+/// Returns the set of fields that are written to via factory.
```
natebosch: blank line after the first sentence

I'm assuming there are side effects to this method (like writing to `buffer`) - maybe we should focus on those in the first description rather than the return value?

json_serializable/lib/src/utils.dart
```diff
@@ -38,5 +51,182 @@ String friendlyNameForElement(Element element) {
...
+Set<FieldElement> writeFactory(
+    StringBuffer buffer,
+    ClassElement classElement,
+    Map<String, FieldElement> fields,
+    String ignoreReason(String fieldName),
+    String deserializeForField(FieldElement field,
+        {ParameterElement ctorParam})) {
+  var fieldsSetByFactory = new Set<FieldElement>();
+  var className = classElement.displayName;
+  // Create the factory method
```
natebosch: what is this meant to be describing?

json_serializable/lib/src/utils.dart
```diff
@@ -38,5 +51,182 @@ String friendlyNameForElement(Element element) {
...
+
+  // Get the default constructor
```
natebosch: this is a "what" comment - is there a "why" we should be documenting? Or should we drop the comment? Or pull it out into a method named `findDefaultConstructor(...)`?

json_serializable/lib/src/utils.dart
```diff
@@ -38,5 +51,182 @@ String friendlyNameForElement(Element element) {
...
+  }
+
+  // find fields that aren't already set by the constructor and that aren't final
+  var remainingFieldsForFactoryBody = fields.values
+      .where((field) => !field.isFinal)
+      .toSet()
+      .difference(fieldsSetByFactory);
+
+  //
+  // Generate the static factory method
```
natebosch: This method is ~100 lines and big section comments like this make me think it could be broken up into different methods.

kevmoo:
@natebosch – fixed copyright year...

json_serializable/lib/src/utils.dart
kevmoo: Removed – old noise.

json_serializable/lib/src/utils.dart
kevmoo: Removed and adding a TODO around https://github.com/dart-lang/json_serializable/issues/50

json_serializable/lib/src/utils.dart
kevmoo: Did a big clean-up here

json_serializable/lib/src/utils.dart
kevmoo: I did a bit of cleanup. There is a lot of state in a few `var`s here. I'd have to create another class to ship them between functions, which seems like more noise than it's worth

PR #125, opened 2018-04-02
Centralize and improve String literal handling

kevmoo:


json_serializable/lib/src/utils.dart
```diff
@@ -11,6 +11,42 @@ import 'package:analyzer/src/dart/resolver/inheritance_manager.dart'
 
 import 'package:source_gen/source_gen.dart';
 
+// TODO: still need handle tripe singe/double quotes within `value`
```
natebosch: typo in triple

json_serializable/lib/src/utils.dart
```diff
@@ -11,6 +11,42 @@ import 'package:analyzer/src/dart/resolver/inheritance_manager.dart'
...
+String escapeDartString(String value) {
```
natebosch: Doc comment?

json_serializable/lib/src/utils.dart
```diff
@@ -11,6 +11,42 @@ import 'package:analyzer/src/dart/resolver/inheritance_manager.dart'
...
+  if (value.contains('\n')) {
+    return "r'''\n$value'''";
```
natebosch: why the leading `/n`?

json_serializable/lib/src/utils.dart
```diff
@@ -11,6 +11,42 @@ import 'package:analyzer/src/dart/resolver/inheritance_manager.dart'
...
+  }
+
+  var containsSingleQuote = value.contains("'");
+  var contains$ = value.contains(r'$');
+
+  if (containsSingleQuote) {
+    if (value.contains('"')) {
+      // `value` contains both single and double quotes as well as `$`.
```
natebosch: we don't know if it contains `$` here. Are we assuming that it could so may as well escape it?

json_serializable/lib/src/utils.dart
```diff
@@ -11,6 +11,42 @@ import 'package:analyzer/src/dart/resolver/inheritance_manager.dart'
...
+      // The only safe way to wrap the content is to escape all of the
+      // problematic characters.
+      var string = value
+          .replaceAll('\$', '\\\$')
```
natebosch: Inconsistent between `'\$'` and `r'$'` from line 21. How about `.replaceAll(r'$', r'\$')`

Similarly for the quotes I think it's more readable with `r` than with double slash.

json_serializable/lib/src/utils.dart

json_serializable/lib/src/utils.dart

json_serializable/lib/src/utils.dart
kevmoo: Because one leading `\n` is ignored. So if it's the first character in the string, it's skipped.

json_serializable/lib/src/utils.dart
kevmoo: Oops! Will dig in a bit...

json_serializable/lib/src/utils.dart
lrhn: (Also ignores leading white space if that's all there is before the first line terminator).

kevmoo:
@natebosch – PTAL

json_serializable/lib/src/utils.dart
```diff
@@ -11,6 +11,42 @@ import 'package:analyzer/src/dart/resolver/inheritance_manager.dart'
...
+          .replaceAll('"', '\\"')
+          .replaceAll("'", "\\'");
```
lrhn: Doing three replacements operation, where at least two are known to change the string, may be a little costly on large strings.
On the other hand, it's probably much cheaper than the alternative:
```dart
  var string = value.replaceAll(new RegExp(r"""(?=[$'"])""", r"\");
```
(even if the `RegExp` and replacement function are both made constant).

Consider whether it might be worth it as a special case for large strings.

Also still a problem if the string contains `\$` already.

kevmoo: Will investigate in a follow-up CL. This is strictly better than what we had before...I think...

PR #126, opened 2018-04-03
Better escaping of String literals

kevmoo:


kevmoo:
CC @lrhn

json_serializable/lib/src/utils.dart
```diff
@@ -15,43 +15,61 @@ import 'package:source_gen/source_gen.dart';
...
+      return value;
+    } else if (value == r'$') {
+      hasDollar = true;
+      return value;
+    }
+
+    canBeRaw = false;
     var mapped = _escapeMap[match[0]];
     if (mapped != null) return mapped;
     return _getHexLiteral(match[0]);
```
lrhn: Those three lines could be just:
```dart
return _escapeMap[value] ?? _getHexLiteral(value);
```
I'd even do:
```dart
return _escapeMap.putIfAbsent(value, _getHexLiteral);
```
(and make the map not constant, then you will cache the hex literals the first time you compute them instead of creating a new string each time).

json_serializable/lib/src/utils.dart
```diff
@@ -15,43 +15,61 @@ import 'package:source_gen/source_gen.dart';
...
+      }
+    } else {
+      // quote it with single quotes!
+      return "r'$value'";
+    }
+  }
+
+  // The only safe way to wrap the content is to escape all of the
+  // problematic characters - `$`, `'`, and `"`
+  var string = value.replaceAll(new RegExp(r"""(?=[$'"])"""), r'\');
```
lrhn: Cache the regexp in a final variable, like for `_escapeRegExp`.

json_serializable/lib/src/utils.dart
```diff
@@ -15,43 +15,61 @@ import 'package:source_gen/source_gen.dart';
...
+
   value = value.replaceAllMapped(_escapeRegExp, (match) {
+    var value = match[0];
+    if (value == "'") {
+      hasSingleQuote = true;
+      return value;
+    } else if (value == '"') {
+      hasDoubleQuote = true;
+      return value;
+    } else if (value == r'$') {
```
lrhn: You could treat `\` the same as `$` wrt. "canBeRaw", a string with content `"$\` can be raw as `r'"$\'`.

I can see that it doesn't work with the way this operation currently works - you can't decide later that the string can't be raw, because then you have to go back and fix previous `\` characters, so I guess this is a good enough trade-off since backslashes are rare.

The alternative is to scan the string once first to figure out what it contains, and then fix it up in one second operation.

json_serializable/lib/src/utils.dart
```diff
@@ -15,43 +15,61 @@ import 'package:source_gen/source_gen.dart';
...
+        // quote it with single quotes!
+        return 'r"$value"';
+      }
+    } else {
+      // quote it with single quotes!
+      return "r'$value'";
+    }
+  }
+
+  // The only safe way to wrap the content is to escape all of the
```
lrhn: I take offense with the word "only" :)
You could potentially use tripple-quotes in some cases (you just don't have the information here to know when).

json_serializable/lib/src/utils.dart
```diff
@@ -48,6 +51,28 @@ String escapeDartString(String value) {
...
+  '\n': r'\n', // 0A - new line
+  '\v': r'\v', // 0B - vertical tab
+  '\f': r'\f', // 0C - form feed
+  '\r': r'\r', // 0D - carriage return
+  '\x7F': r'\x7F', // delete
+};
+
+/// A [RegExp] that matches whitespace characters that should be escaped.
+final _escapeRegExp = new RegExp(
+    '[\\x00-\\x07\\x0E-\\x1F${_escapeMap.keys.map(_getHexLiteral).join()}]');
```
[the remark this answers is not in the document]
lrhn: Nevermind, you extended the map in a later change, then it does make sense. I'd still prboably start with just `\\x00-\\x1f' and not mind the overlap.

json_serializable/lib/src/utils.dart
```diff
@@ -15,9 +15,12 @@ import 'package:source_gen/source_gen.dart';
 /// Dart code.
 // TODO: still need handle triple singe/double quotes within `value`
 String escapeDartString(String value) {
-  if (value.contains('\n')) {
-    return "r'''\n$value'''";
-  }
+  value = value.replaceAll('\\', r'\\');
```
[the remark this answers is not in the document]
kevmoo: this is for an old commit, I think...

json_serializable/lib/src/utils.dart
kevmoo: Ooo...not a bad idea...

PR #129, opened 2018-04-03
missing dep on pkg:logging

kevmoo:


kevmoo:
@nex3 – the dependency checker feature of `pub` is soooooo nice!

PR #130, opened 2018-04-03
Update example to latest publish packages

kevmoo:


kevmoo:
@jakemac53 – last one today – I swear!

PR #131, opened 2018-04-05
Generalize an isEnum helper, added to utils.dart

kevmoo:


kevmoo:
One more thing to eventually move into `source_gen`

kevmoo:
@jakemac53 – is nice 😄

PR #134, opened 2018-04-11
fix for prefer_const_constructors lint

kevmoo:


kevmoo:
...oops...should have waited for Travis

PR #158, opened 2018-05-16
Add `any_map` support

kevmoo:


json_serializable/test/config_test.dart
```diff
@@ -0,0 +1,76 @@
...
+  tearDownAll(() async {
+    await sub.cancel();
+  });
+
+  setUp(records.clear);
+
+  test('empty', () async {
+    var builder = jsonSerializable(BuilderOptions.empty) as PartBuilder;
+    expect(builder, isNotNull);
+    expect(records, isEmpty);
```
jakemac53: Note that since the logs come through asynchronously you would never see them at this point, you would need to add an `await null` or something along those lines. Not really sure the absolute best way to deal with it honestly.

json_serializable/test/config_test.dart
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  group('invalid config', () {
+    for (var entry in _invalidConfig.entries) {
+      test(entry.key, () {
+        var config = new Map<String, dynamic>.from(_validConfig);
+        config[entry.key] = entry.value;
+
+        expect(() => jsonSerializable(new BuilderOptions(config)),
+            _throwsCastError);
```
jakemac53: Worth checking the actual message here?

json_serializable/test/config_test.dart
```diff
@@ -0,0 +1,76 @@
...
+    expect(records, isEmpty);
+  });
+
+  test('unsupported configuration', () async {
+    var builder =
+        jsonSerializable(const BuilderOptions(const {'unsupported': 'config'}))
+            as PartBuilder;
+    expect(builder, isNotNull);
+
+    expect(records.single.message,
```
jakemac53: I don't get how this is passing given my comment above...

json_serializable/test/test_files/kitchen_sink.g.dart
```diff
@@ -38,21 +38,22 @@ KitchenSink _$KitchenSinkFromJson(Map<String, dynamic> json) => new KitchenSink(
   ..stringDateTimeMap = json['stringDateTimeMap'] == null
       ? null
       : new Map<String, DateTime>.fromIterables(
-          (json['stringDateTimeMap'] as Map<String, dynamic>).keys,
+          (json['stringDateTimeMap'] as Map).keys.cast<String>(),
```
jakemac53: nit: should we be using `.map` instead?

json_serializable/test/config_test.dart
kevmoo: I don't think so. They will all be 'String is not a bool' etc.

json_serializable/test/config_test.dart
kevmoo: as discussed...

json_serializable/test/test_files/kitchen_sink.g.dart
kevmoo: Same-same, right? This seems easier...

json_serializable/test/config_test.dart
jakemac53: Ya I was just wondering if you wanted to check the cast messages were correct, but its probably overkill

PR #169, opened 2018-05-22
Add support for generic classes

kevmoo:
...and some name cleanup.

I will land these are separate commits, @natebosch 😄

### DIFF
--- a/json_serializable/lib/src/generator_helper.dart
+++ b/json_serializable/lib/src/generator_helper.dart
@@ -39,8 +39,22 @@ class _GeneratorHelper {
   final StringBuffer _buffer = new StringBuffer();
 
   String get _prefix => '_\$${_element.name}';
-  String get _mixClassName => '${_prefix}SerializerMixin';
-  String get _helpClassName => '${_prefix}JsonMapWrapper';
+
+  String _mixinClassName(bool withConstraints) =>
+      '${_prefix}SerializerMixin${_genericClassArguments(withConstraints)}';
+
+  String _wrapperClassName([bool withConstraints]) =>
+      '${_prefix}JsonMapWrapper${_genericClassArguments(withConstraints)}';
+
+  /// Returns a [String] representing the type arguments that exist on
+  /// [_element].
+  ///
+  /// Returns the output of calling [genericClassArguments] with [_element].
+  String _genericClassArguments(bool withConstraints) =>
+      genericClassArguments(_element, withConstraints);
+
+  String get _targetClassReference =>
+      '${_element.name}${_genericClassArguments(false)}';
 
   _GeneratorHelper(this._generator, this._element, this._annotation);
 
@@ -66,7 +80,7 @@ class _GeneratorHelper {
       //
       // Generate the mixin class
       //
-      _buffer.writeln('abstract class $_mixClassName {');
+      _buffer.writeln('abstract class ${_mixinClassName(true)} {');
 
       // write copies of the fields - this allows the toJson method to access
       // the fields of the target class
@@ -80,7 +94,7 @@ class _GeneratorHelper {
       var writeNaive = accessibleFields.every(_writeJsonValueNaive);
 
       if (_generator.useWrappers) {
-        _buffer.writeln('=> new $_helpClassName(this);');
+        _buffer.writeln('=> new ${_wrapperClassName(false)}(this);');
       } else {
         if (writeNaive) {
           // write simple `toJson` method that includes all keys...
@@ -126,7 +140,8 @@ class _GeneratorHelper {
     if (_annotation.createFactory) {
       _buffer.writeln();
       var mapType = _generator.anyMap ? 'Map' : 'Map<String, dynamic>';
-      _buffer.writeln('${_element.name} ${_prefix}FromJson($mapType json) =>');
+      _buffer.writeln('$_targetClassReference '
+          '${_prefix}FromJson${_genericClassArguments(true)}($mapType json) =>');
 
       String deserializeFun(String paramOrFieldName,
               {ParameterElement ctorParam}) =>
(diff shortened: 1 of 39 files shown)


natebosch:
> I will land these are separate commits,

Why couldn't they be separate PRs?

natebosch:
Also, why does a commit which adds over 400 lines have a 5 word commit message?

json_serializable/lib/src/generator_helper.dart
```diff
@@ -39,8 +39,20 @@ class _GeneratorHelper {
...
 
   String get _prefix => '_\$${_element.name}';
-  String get _mixClassName => '${_prefix}SerializerMixin';
-  String get _helpClassName => '${_prefix}JsonMapWrapper';
+  String _mixinClassName(bool withConstraints) =>
+      '${_prefix}SerializerMixin${_genericClassArguments(withConstraints)}';
+  String _wrapperClassName([bool withConstraints]) =>
+      '${_prefix}JsonMapWrapper${_genericClassArguments(withConstraints)}';
+
+  String _genericClassArguments(bool withConstraints) {
```
natebosch: Doc comment?

json_serializable/test/generic_files/generic_class.dart
```diff
@@ -0,0 +1,41 @@
...
+// ignore_for_file: annotate_overrides
+
+import 'package:json_annotation/json_annotation.dart';
+
+part 'generic_class.g.dart';
+
+@JsonSerializable()
+class GenericClass<T extends num, S> extends Object
+    with _$GenericClassSerializerMixin<T, S> {
+  @JsonKey(fromJson: _dataFromJson, toJson: _dataToJson)
```
jakemac53: are `_dataFromJson` and `_dataToJson` required for this to work?

json_serializable/lib/src/generator_helper.dart
```diff
@@ -39,8 +39,20 @@ class _GeneratorHelper {
   final StringBuffer _buffer = new StringBuffer();
 
   String get _prefix => '_\$${_element.name}';
-  String get _mixClassName => '${_prefix}SerializerMixin';
-  String get _helpClassName => '${_prefix}JsonMapWrapper';
+  String _mixinClassName(bool withConstraints) =>
```
natebosch: consider making this 2 separate methods (or getters) rather than using a flag argument. The names will be more clear about what is happening than a cryptic "true" or "false" for which you need to see the argument name to understand.

json_serializable/lib/src/generator_helper.dart
```diff
@@ -221,9 +234,10 @@ class _GeneratorHelper {
   void _writeWrapper(Iterable<FieldElement> fields) {
     _buffer.writeln();
     // TODO(kevmoo): write JsonMapWrapper if annotation lib is prefix-imported
-    _buffer.writeln('''class $_helpClassName extends \$JsonMapWrapper {
-      final $_mixClassName _v;
-      $_helpClassName(this._v);
+    _buffer
+        .writeln('''class ${_wrapperClassName(true)} extends \$JsonMapWrapper {
```
natebosch: with `writeln` you're going to get an extra newline and this already ends in a newline. Do we need to?

json_serializable/lib/src/utils.dart
```diff
@@ -175,7 +175,19 @@ int _sortByLocation(FieldElement a, FieldElement b) {
 
 final _dartCoreObjectChecker = const TypeChecker.fromRuntime(Object);
 
-/// If a parameter is required to invoke the constructor,
+String genericClassArguments(ClassElement element, bool withConstraints) {
```
natebosch: we seem to be relying on `withConstraints` being non-null here because of  another method with the same name handling nullability. Maybe handle them both here and skip the second identically named wrapper method?

natebosch: Doc comment?

json_serializable/lib/src/utils.dart
```diff
@@ -175,7 +175,19 @@ int _sortByLocation(FieldElement a, FieldElement b) {
...
+  if (element.typeParameters.isEmpty) {
+    return '';
+  }
+  var values = element.typeParameters.map((t) {
+    if (withConstraints) {
+      return t.toString();
```
natebosch: consider ternary conditional when what you care about is the value rather than side effects

json_serializable/lib/src/utils.dart
```diff
@@ -243,7 +255,10 @@ CtorData writeConstructorInvocation(
       writeableFields.toSet().difference(usedCtorParamsAndFields);
 
   var buffer = new StringBuffer();
-  buffer.write('new $className(');
+  //
+  // Generate the static factory method
```
natebosch: should this be a separate method named `generateStaticFactory`?

json_serializable/test/generic_files/generic_class.dart
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
natebosch: 2018

json_serializable/test/generic_files/generic_class.dart
kevmoo: Yup. Will document as such.

json_serializable/lib/src/generator_helper.dart
kevmoo: ack

json_serializable/lib/src/generator_helper.dart
kevmoo: Honestly, dartfmt handles most of this. Not going to modify this in this CL

json_serializable/lib/src/utils.dart
kevmoo: Ack. Going to leave both. Having the helper in `generator_helper` is nice.

Did forward the null check to the inner method

json_serializable/lib/src/utils.dart

json_serializable/lib/src/utils.dart
kevmoo: This is an old comment. This does not create a factory. It creates an invocation. Will remove.

json_serializable/test/generic_files/generic_class.dart
kevmoo: ack

json_serializable/test/generic_files/generic_class.dart
kevmoo: https://github.com/dart-lang/json_serializable/issues/171

PR #170, opened 2018-05-22
Standardize on Element.name everywhere

kevmoo:


kevmoo:
@natebosch separate PR 😄

PR #173, opened 2018-05-22
annotation: Update library doc comment describing the `$` helper usage

kevmoo:
Also finish documenting the json_serializable annotations

### DIFF
--- a/json_annotation/lib/src/json_serializable.dart
+++ b/json_annotation/lib/src/json_serializable.dart
@@ -4,8 +4,32 @@
 
 /// An annotation used to specify a class to generate code for.
 class JsonSerializable {
-  // TODO(kevmoo): document these fields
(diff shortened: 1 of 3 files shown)


json_annotation/lib/src/json_serializable.dart
```diff
@@ -22,7 +46,10 @@ class JsonSerializable {
   /// `null` runtime validation if it's critical.
   final bool nullable;
 
-  // TODO(kevmoo): document the constructor
+  /// Creates a [JsonSerializable] instance.
+  ///
+  /// For details on each parameter, see the documentation for the corresponding
```
natebosch: [nit] do we think this is something that we need to state explicitly? I don't think I've ever hesitated to jump to field doc comments.

PR #181, opened 2018-05-24
Fix bug with enum closure parameter

kevmoo:


json_serializable/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.5.5
+## 0.5.4+1
 
 * Added named `formatOutput` parameter to `jsonPartBuilder`.
```
natebosch: adding a parameter (A feature someone can depend on) should be done in a minor release, not a bug fix release. The previous version in this file was better.

json_serializable/lib/src/type_helpers/enum_helper.dart
```diff
@@ -34,6 +34,11 @@ class EnumHelper extends TypeHelper {
     var wrappedExpression =
         simpleExpression.hasMatch(expression) ? expression : '{$expression}';
 
+    var closureArg = consts.closureArg;
```
natebosch: so was the import unused before? How is it that you could add a prefix without changing any other lines?

kevmoo: It wasn't unused before. I'm creating a local var that maps exactly to the field `closureArg`. This way the code that does the generation doesn't change at all...

json_serializable/CHANGELOG.md
kevmoo: Er, uh. Yup. Will fix.

...although the named arg will be useful to very few people. But fair...

kevmoo: Went the other way and removed the changelog entry – jsonPartBuilder is not public!

PR #182, opened 2018-05-25
Starting on default value work

kevmoo:
I'd like to do a bit more with checking types and supporting enums, but this is a good start


json_serializable/test/default_value/default_value_test.dart
```diff
@@ -0,0 +1,51 @@
...
+
+const _otherValues = const {
+  'fieldBool': false,
+  'fieldString': 'other string',
+  'fieldInt': 43,
+  'fieldDouble': 2.718,
+  'fieldListEmpty': const [42],
+  'fieldMapEmpty': const {'question': false},
+  'fieldListSimple': const [4, 5, 6],
+  'fieldMapSimple': const <Null, Null>{},
```
jakemac53: `<String, int>`?

kevmoo:
@jakemac53 – approved otherwise?

PR #193, opened 2018-05-28
Fix detection of `toJson`/`fromJson` in nested types

kevmoo:


json_serializable/lib/src/type_helpers/json_helper.dart
```diff
@@ -51,17 +65,55 @@ class JsonHelper extends TypeHelper {
...
+      // TODO: consider logging that we're assuming a user will wire up the
+      // generated mixin at some point...
       return true;
     }
   }
-
   return false;
 }
+
+JsonSerializable _getAnnotation(InterfaceType source) {
```
natebosch: [nit] how about `annotation`.

https://www.dartlang.org/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

kevmoo: fixed

kevmoo:
Updated commit message and rebased. PTAL @natebosch

PR #195, opened 2018-05-28
Add option for `implicitToJson`

### LINKED ISSUE #192: Add an option to explicitly call toJson when generating serialization code
So rummaging through some older issues, I found that the expected way to handle an structure like:

```
User
---List<Favorites>
```
Is to do something like this:

`Map<String, dynamic> jsonObject = json.decode(json.encode(user)); //json is from dart:convert`

Instead of the normal user.toJson which works with objects without other nested objects inside of  them. This isn't a huge deal, but it just makes the code look a little weird. If there's a better way to do this please let me know. Thanks!

kevmoo:
Allows users to opt-in to including `.toJson()` calls on nested
objects in generated `toJson` methods.

This is useful if you want to round-trip the output of toJson without
putting it through jsonEncode and jsonDecode

Fixes #192

### DIFF
--- a/json_serializable/lib/src/type_helpers/json_helper.dart
+++ b/json_serializable/lib/src/type_helpers/json_helper.dart
@@ -21,11 +21,15 @@ class JsonHelper extends TypeHelper {
   /// By default, JSON encoding in from `dart:convert` calls `toJson()` on
   /// provided objects.
   @override
-  String serialize(DartType targetType, String expression, _) {
+  String serialize(
+      DartType targetType, String expression, SerializeContext context) {
     if (!_canSerialize(targetType)) {
       return null;
     }
 
+    if (context is TypeHelperContext && context.explicitToJson) {
+      return '$expression${context.nullable ? '?' : ''}.toJson()';
+    }
     return expression;
   }
 
--- a/json_serializable/lib/src/type_helpers/json_helper.dart
+++ b/json_serializable/lib/src/type_helpers/json_helper.dart
@@ -16,10 +16,10 @@ import '../utils.dart';
 class JsonHelper extends TypeHelper {
   const JsonHelper();
 
-  /// Simply returns the [expression] provided.
+  /// If the builder is configured with `implicit_to_json: true`, simply returns
+  /// the [expression] provided.
   ///
-  /// By default, JSON encoding in from `dart:convert` calls `toJson()` on
-  /// provided objects.
+  /// Otherwise, returns [expression]`.toJson()` with proper `null` handling.
   @override
   String serialize(
       DartType targetType, String expression, SerializeContext context) {
@@ -27,10 +27,13 @@ class JsonHelper extends TypeHelper {
       return null;
     }
 
-    if (context is TypeHelperContext && context.explicitToJson) {
-      return '$expression${context.nullable ? '?' : ''}.toJson()';
+    var ctx = context as TypeHelperContext;
+
+    if (ctx.implicitToJson) {
+      return expression;
     }
-    return expression;
+
+    return '$expression${context.nullable ? '?' : ''}.toJson()';
   }
 
   @override
--- a/json_serializable/test/src/json_serializable_test_input.dart
+++ b/json_serializable/test/src/json_serializable_test_input.dart
@@ -214,3 +214,13 @@ class SuperType {
   int priceFraction(int other) =>
       superTypeViaCtor == null ? null : superTypeViaCtor ~/ other;
 }
+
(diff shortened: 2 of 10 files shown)


kevmoo:
CC @jmrboosties – let me know if this makes sense...

jmrboosties:
Looks great! Thanks a lot.

natebosch:
I'm not sure about the naming... I don't think this is about "explicit" vs "implicit". Maybe something like `nestedToJson`?

Also, is there any reason not do do this by default?

kevmoo:
> I'm not sure about the naming... I don't think this is about "explicit" vs "implicit". Maybe something like nestedToJson?

I was thinking that the call to `toJson` is implicit in `dart:convert`. This makes it explicit. Not sure `nestedToJson` is any more clear...

> Also, is there any reason not do do this by default?

I pondered making this `implicitToJson` with a default of `true` (unlike the other options) and a note that we'll flip the default in the next release.

I think it makes the behavior a bit more clear...

json_serializable/lib/src/type_helpers/json_helper.dart
```diff
@@ -21,11 +21,15 @@ class JsonHelper extends TypeHelper {
   /// By default, JSON encoding in from `dart:convert` calls `toJson()` on
```
jakemac53: nit: update this comment

json_serializable/test/src/json_serializable_test_input.dart
```diff
@@ -214,3 +214,13 @@ class SuperType {
   int priceFraction(int other) =>
       superTypeViaCtor == null ? null : superTypeViaCtor ~/ other;
 }
+
+@JsonSerializable(createFactory: false)
+class TrivialNestedNullable {
+  TrivialNestedNullable child;
```
jakemac53: consider adding an `int` or some other non-custom type here, we want to make sure we don't call `toJson` on those

json_serializable/lib/src/type_helpers/json_helper.dart
kevmoo: how?

json_serializable/test/src/json_serializable_test_input.dart
kevmoo: sure...the code paths for this are so crazy independent, I'm not too worried – but paranoia is good!

json_serializable/lib/src/type_helpers/json_helper.dart
jakemac53: By noting that if `context.explicitToJson` is `true` these calls will be injected automatically.

json_serializable/test/src/json_serializable_test_input.dart
jakemac53: It also prevents future regressions etc, looking at this PR alone its not clear this has the correct behavior today.

kevmoo: done

jakemac53:
I like `explicitToJson`, and I think eventually defaulting it to true I think. These things should be usable outside of the built in json encoding/decoding.

I actually ran into this with the firebase package if I recall... I think it doesn't use the built in json encode/decode functionality.

json_serializable/lib/src/type_helpers/json_helper.dart
kevmoo: done

kevmoo:
Name/default flipped. PTAL @jakemac53 @natebosch

natebosch:
I still don't think "implicit" vs "explicit" is the distinction this is making. This is about an output which is _only_ compatible with `jsonEncode` vs being valid json types. "shallow" vs "nested" is the distinction this is making. With a "shallowToJson" the nested fields are _not_ converted to json types, with a "nestedToJson" the nested fields _are_ converted to json types.

jakemac53:
What about `recursiveToJson`?

natebosch:
The other option is to skip the option, make it always on, and do the breaking version bump with this change. This package has a lot of configuration options already and it might not hurt to reduce the surface area a bit.

kevmoo:
> The other option is to skip the option, make it always on, and do the breaking version bump with this change. This package has a lot of configuration options already and it might not hurt to reduce the surface area a bit.

1) I'd rather hold off on this until the next release. We have a lot of stuff queued up.

2) A nice thing we throw away here: checks for loops. `dart:convert` does the work to make sure we only visit `Customer: hash 1234abcd` once.

kevmoo:
Going to hold off on this PR for a bit...

PR #225, opened 2018-06-06
DO NOT SUMBIT: Question: is this a better way to get at enum values via pkg:analyzer?

kevmoo:
@scheglov @bwilkerson as you can likely tell be recent issues ( https://github.com/dart-lang/sdk/issues/33373 https://github.com/dart-lang/sdk/issues/33375 ) I'm looking at some code that details with enums.

Look at the diff here. Both before and after have their own code smell – is one preferred? Is there something I'm missing?

Thanks!

### DIFF
--- a/json_serializable/lib/src/json_key_with_conversion.dart
+++ b/json_serializable/lib/src/json_key_with_conversion.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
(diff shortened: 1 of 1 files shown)


bwilkerson:
Neither solution is great, but there isn't anything better that I can think of. The element model doesn't differentiate between classes and enums, so there isn't any nice way to get all of the enum constants.

When analyzer moves to the CFE we might not be able to move as easily from the element model back to the AST, so the current code might be slightly better in terms of future-proofing.

One alternative I can think of would be to ask for `fields` rather than `accessors`, and then ask whether the field is synthetic. (Everything except the explicitly defined enum constants will be synthetic.) It might be a little more performant because we don't have to test the equality of types, but is fairly equivalent in other ways.

kevmoo:
Thanks, @bwilkerson – opened https://github.com/dart-lang/json_serializable/pull/227

Would love comments on the other issues I opened.

PR #228, opened 2018-06-07
Add option for `explicitToJson`

### LINKED ISSUE #192: Add an option to explicitly call toJson when generating serialization code

kevmoo:
Allows users to opt-in to including `.toJson()` calls on nested
objects in generated `toJson` methods.

This is useful if you want to round-trip the output of toJson without
putting it through jsonEncode and jsonDecode

Fixes #192

### DIFF
--- a/json_serializable/CHANGELOG.md
+++ b/json_serializable/CHANGELOG.md
@@ -17,6 +17,8 @@
 * Added missing `checked` parameter to the
   `JsonSerializableGenerator.withDefaultHelpers` constructor.
 
+* Added `explicit_to_json` configuration option.
+
 ## 0.5.6
 
 * Added support for `JsonSerializable.disallowUnrecognizedKeys`.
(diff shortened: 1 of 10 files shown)


kevmoo:
Since `dart:convert` does the work to ensure there are no cycles, I'm going to leave this opt-in for users.

kevmoo:
See discussion in previous PR - https://github.com/dart-lang/json_serializable/pull/195

PR #230, opened 2018-06-08
Added new generate_to_json_function build option

kevmoo:
Controls how `toJson` functionality is generated for all types processed
by this generator.

If `false` (the default), a private `_$ClassNameSerializerMixin` class is
created in the generated part file which contains a `toJson` method.

Mix in this class to the source class:

```dart
@JsonSerializable()
class Example extends Object with _$ExampleSerializerMixin {
  // ...
}
```

If `true`, then a top-level function is created that you can reference
from your class.

```dart
@JsonSerializable()
class Example {
  // ...
  Map<String, dynamic> toJson() => _$ExampleToJson(this);
}
```

### DIFF
--- a/json_serializable/lib/src/generator_helper.dart
+++ b/json_serializable/lib/src/generator_helper.dart
@@ -33,6 +33,10 @@ Future<String> generate(JsonSerializableGenerator generator, Element element,
 }
 
 class _GeneratorHelper {
+  /// Name of the parameter used when generating top-level `toJson` functions
+  /// if [JsonSerializableGenerator.generateToJsonFunction] is `true`.
+  static const _toJsonParamName = 'instance';
+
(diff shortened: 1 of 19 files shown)


json_serializable/lib/src/generator_helper.dart
```diff
@@ -274,12 +243,67 @@ class _GeneratorHelper {
...
+    }
+
+    if (_generator.generateToJsonFunction) {
+      var functionName = '${_prefix}ToJson${_genericClassArguments(true)}';
+      _buffer.write('Map<String, dynamic> $functionName'
+          '($_targetClassReference $_toJsonParamName) ');
+    } else {
+      //
+      // Generate the mixin class
+      //
```
jakemac53: :D?

PR #235, opened 2018-06-12
misc: enable and fix a number of lints

kevmoo:


json_serializable/tool/build.dart
```diff
@@ -118,6 +119,6 @@ final List<BuilderApplication> builders = [
       hideOutput: true),
 ];
 
-main(List<String> args) async {
+Future main(List<String> args) async {
```
natebosch: [optional] could be `void`.

kevmoo: TIL - nice!

PR #240, opened 2018-06-18
refactor: prepare for multi-return values in GenerateForAnnotation

kevmoo:
JsonSerializableGenerator now returns String instead of Future<String>
GeneratorHelper is now private and in the same file is the Generator
Encoding and decoding have been split into mixin classes
  These two aspects of the generator are nicely isolated hopefully
  making future fixes and additions easier

Many members of utils.dart have been moved closer to where they are used
  and made private when possible

Updated tests to be sync now that the core tested function is sync


json_serializable/lib/src/decode_helper.dart
```diff
@@ -0,0 +1,302 @@
+import 'package:analyzer/dart/element/element.dart';
+import 'package:json_annotation/json_annotation.dart';
+import 'package:source_gen/source_gen.dart';
+
+import 'helper_core.dart';
+import 'json_literal_generator.dart';
+import 'type_helper.dart';
+import 'utils.dart';
+
+abstract class DencodeHelper implements HelperCore {
```
jakemac53: nit: `DecodeHelper`

json_serializable/lib/src/decode_helper.dart
```diff
@@ -0,0 +1,302 @@
+import 'package:analyzer/dart/element/element.dart';
```
jakemac53: copyright

kevmoo:
> Did not re-review internals, this is mostly just moves right? 

Yup. Mostly copy-paste. The mixin thing is pretty nice...

jakemac53:
Ya, I like the new structure more for sure

Issue #233, opened 2018-06-08
CheckedFromJsonException : Get message out of FormatException, too


kevmoo:
https://github.com/dart-lang/json_serializable/blob/95dc1175eb8dba89591f8ca5ac6faa8be8a44838/json_annotation/lib/src/checked_helpers.dart#L94-L101

PR #242, opened 2018-06-20
Fix for #233 and other things

kevmoo:
Will land separately


json_serializable/test/yaml/yaml_test.dart
```diff
@@ -171,7 +179,7 @@ String _prettyPrintCheckedFromJsonException(CheckedFromJsonException err) {
     var yamlValue = yamlMap.nodes[err.key];
 
     if (yamlValue == null) {
-      assert(err.key == null);
+      //assert(err.key == null);
```
jakemac53: ?

PR #243, opened 2018-06-20
feature: allow undefined types as long as they have a converter

kevmoo:
Types from Platform-specific libraries (like dart:ui in Flutter)
don't resolve properly during code generation.
See https://github.com/dart-lang/build/issues/733

This update allows undefined types to be handled with custom
to/fromJson for an associated field

Fixes https://github.com/dart-lang/json_serializable/issues/236

### DIFF
--- a/json_serializable/test/src/json_serializable_test_input.dart
+++ b/json_serializable/test/src/json_serializable_test_input.dart
@@ -105,6 +105,21 @@ class UnknownFieldType {
 }
 
 @JsonSerializable(createFactory: false)
+class UnknownFieldTypeToJsonOnly {
+  // ignore: undefined_class
+  Bob number;
+}
+
+@JsonSerializable()
+class UnknownFieldTypeWithConvert {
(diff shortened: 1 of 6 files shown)


json_serializable/test/src/json_serializable_test_input.dart
```diff
@@ -104,6 +104,21 @@ class UnknownFieldType {
   Bob number;
 }
 
+@JsonSerializable(createFactory: false)
+class UnknownFieldTypeToJsonOnly {
+  // ignore: undefined_class
+  Bob number;
```
jakemac53: Note that in Dart 2 this will be a compile time error so migrating this test to use modular kernel will likely be an issue, but you could cross that bridge when you come to it (or possibly exclude this file from that builder).

kevmoo: ack – it was either this or depend on Flutter in the Test 😄

PR #251, opened 2018-06-28
feature: allow customizing enum value serialization via annotations

kevmoo:
feature: allow customizing enum value serialization via annotations

* Require the latest Dart SDK
  Annotated enums fail without the CFE

* Require the latest pkg:analyzer
  Enum metadata parsing was just recently added

* Require the latest pkg:source_gen
  Uses multiple return values to allow generating shared helpers

* Removed several helpers in json_annotation
  They can now be generated as part of the source, which minimizes
  versioning issues with helpers.

* Small, but breaking changes to Generator API to return multiple
  values

* Added `addMember` API to Context classes to allow helpers to be
  generated.

* Updated integration test code to include an annotated enum

* Updated yaml test code to remove now superfluous convert logic and use
  annotated enums instead

Fixes https://github.com/dart-lang/json_serializable/issues/38

### DIFF
--- a/json_annotation/lib/src/json_serializable.dart
+++ b/json_annotation/lib/src/json_serializable.dart
@@ -200,5 +200,6 @@ class JsonValue {
   ///
   /// Can be a [String] or an [int].
   final dynamic value;
+
   const JsonValue(this.value);
 }
--- a/json_annotation/lib/src/json_serializable.dart
+++ b/json_annotation/lib/src/json_serializable.dart
@@ -194,44 +194,11 @@ class JsonKey {
   });
 }
 
-// Until enum supports parse: github.com/dart-lang/sdk/issues/33244
-/// *Helper function used in generated code with `enum` values – should not be
-/// used directly.*
-///
-/// Returns an enum instance corresponding to [enumValue] from the enum named
-/// [enumName] with [values].
-///
-/// If [enumValue] is null or no corresponding values exists, an `ArgumentError`
-/// is thrown.
-///
-/// Given an enum named `Example`, an invocation would look like
-///
-/// ```dart
-/// $enumDecode('Example', Example.values, 'desiredValue')
-/// ```
-T $enumDecode<T>(String enumName, List<T> values, String enumValue) =>
-    values.singleWhere((e) => e.toString() == '$enumName.$enumValue',
-        orElse: () => throw new ArgumentError(
-            '`$enumValue` is not one of the supported values: '
-            '${values.map(_nameForEnumValue).join(', ')}'));
-
-/// *Helper function used in generated code with `enum` values – should not be
-/// used directly.*
-///
-/// Returns an enum instance corresponding to [enumValue] from the enum named
-/// [enumName] with [values].
-///
-/// If [enumValue] is `null`, `null` is returned.
-///
-/// If no corresponding values exists, an `ArgumentError` is thrown.
-///
-/// Given an enum named `Example`, an invocation would look like
-///
-/// ```dart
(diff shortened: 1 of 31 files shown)


kevmoo:
Question for reviewers (@natebosch , @jakemac53 )

Should I create another annotation class – `JsonEnumValue` or similar?
  - Only using `name` field from `JsonKey` others are ignored
  - Would allow num values (as asked https://github.com/dart-lang/json_serializable/issues/38#issuecomment-400503308)

sir-boformer:
Just to give some reasoning: Integers are the default representation of enums in many other languages (e.g. C#, Kotlin, TypeScript), and it's also common to serialize enum values as integers, as it is a very efficient format.

kevmoo:
@sir-boformer – added `JsonValue` class which supports String and int values.

@jacob314 – ready for review

json_serializable/lib/src/type_helpers/enum_helper.dart
```diff
@@ -15,26 +16,75 @@ class EnumHelper extends TypeHelper {
...
   String deserialize(
       DartType targetType, String expression, DeserializeContext context) {
-    if (!isEnum(targetType)) {
+    var memberContent = _enumValueMapFromType(targetType);
+
+    if (memberContent == null) {
       return null;
     }
 
+    context.addMember(_enumDecodeHelper);
```
jakemac53: So we will actually be adding one of these each time this helper is called right? Is that once per field using the enum or more than that?

json_serializable/lib/src/utils.dart
```diff
@@ -27,6 +27,62 @@ JsonSerializable valueForAnnotation(ConstantReader annotation) =>
...
+/// the serialized output.
+///
+/// By default, the [String] value is just the name of the enum value.
+/// If the enum value is annotated with [JsonKey], then the `name` property is
+/// used if it's set and not `null`.
+///
+/// If [targetType] is not an enum, `null` is returned.
+Map<FieldElement, dynamic> enumFieldsMap(DartType targetType) {
+  if (targetType is InterfaceType && targetType.element.isEnum) {
+    var value = _enumMapExpando[targetType];
```
jakemac53: nit: `return _enumMapExpando[targetType] ??= ...`?

jakemac53: also, I would consider taking moving big anonymous closure out to a private method or even just a local variable so its easier to reason about what is happening here

json_serializable/lib/src/type_helpers/enum_helper.dart
kevmoo: once per field using the enum – but then they are collapsed.

The storage is a set, so it's only recorded once.

jakemac53: kk

json_annotation/lib/src/json_serializable.dart
```diff
@@ -194,44 +194,11 @@ class JsonKey {
...
-
-// Until enum has a name property: github.com/dart-lang/sdk/issues/21712
-String _nameForEnumValue(Object value) => value.toString().split('.')[1];
+/// An annotation used to specify how a enum value is serialized.
+class JsonValue {
+  /// The value to use when serializing and deserializing.
+  ///
+  /// Can be a [String] or an [int].
+  final dynamic value;
+  const JsonValue(this.value);
```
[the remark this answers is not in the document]
kevmoo: ack

json_serializable/lib/src/utils.dart

kevmoo:
@jakemac53 – PTAL

PR #254, opened 2018-07-09
feature: default to creating toJson function instead of a mixin

kevmoo:
The default for the `generate_to_json_function` is now `true`.

### DIFF
--- a/json_serializable/CHANGELOG.md
+++ b/json_serializable/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 0.6.0
 
+* **BREAKING** By default, code generated to support `toJson` now creates
+  a top-level function instead of a mixin. The default for the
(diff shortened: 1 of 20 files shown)


json_serializable/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## 0.6.0
 
+* **BREAKING** By default, code generated to support `toJson` now creates
+  a top-level function instead of a mixin.
+  * The default for the `generate_to_json_function` is now `true`.
+  * To opt-out of this change, set `generate_to_json_function` to `false`.
```
natebosch: do these need to be separate sub bullets? They seem like a continuation of the same thought.

kevmoo:
Happy to make it a sentence.

PR #263, opened 2018-07-20
Merge 0_5_x into master

kevmoo:


.travis.yml
```diff
@@ -60,7 +60,7 @@ stages:
 
 # Only building master means that we don't run two builds for each pull request.
 branches:
-  only: [master]
+  only: [master,0_5_x]
```
natebosch: could skip this part?

kevmoo: Going to leave it until we publish 0.6...

PR #265, opened 2018-07-22
test: move `expectThrow` values to annotation on target elements

kevmoo:
For json_serializable_test - makes things much more DRY


jakemac53:
This also makes the test more opaque and brittle, i would at least add a check that you found all the annotations you were expecting to find?

kevmoo:
> This also makes the test more opaque and brittle, i would at least add a check that you found all the annotations you were expecting to find?

As I went through these, I made sure that everything was caught.

Uh...I could have a list of all of the classes I expect as another test?

PR #281, opened 2018-08-09
WIP: I279 snake kebab

kevmoo:


json_serializable/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 1.1.0
 
+* Added support for automatically converting field names to JSON map keys as
+  `kebab-case` or `snake_case` with a new option on the `JsonSerializable`
+  annotation. 
+
+## 1.0.1
```
kevmoo: Oops from previous deploy!

kevmoo:
This also prepares a release of `json_annotation`

json_serializable/test/analysis_utils.dart
```diff
@@ -3,20 +3,26 @@
...
 import 'package:source_gen/source_gen.dart';
 
-Future<LibraryReader> resolveCompilationUnit(String path) async {
-  var contextLocator = ContextLocator();
-  var roots = contextLocator.locateRoots(includedPaths: [path]);
-  if (roots.length != 1) {
-    throw StateError('Expected exactly one context root, got $roots');
-  }
+Future<LibraryReader> resolveCompilationUnit(String sourceDirectory) async {
+  var files = Directory(sourceDirectory).listSync().whereType<File>().toList();
```
kevmoo: I wish we had https://github.com/dart-lang/build/issues/1744

json_serializable/test/analysis_utils.dart
```diff
@@ -3,20 +3,26 @@
...
 
-  var analysisContext =
-      ContextBuilder().createContext(contextRoot: roots.single);
-  var resolveResult = await analysisContext.currentSession.getResolvedAst(path);
-  return LibraryReader(resolveResult.unit.element.library);
+  // Sort files to ensure the "first" one is first
```
natebosch: what does this mean?

json_serializable/lib/src/json_key_with_conversion.dart
```diff
@@ -157,6 +158,25 @@ class JsonKeyWithConversion extends JsonKey {
     assert(!this.includeIfNull || !this.disallowNullValue);
   }
 
+  static String _processName(JsonSerializable classAnnotation,
+      String jsonKeyNameValue, FieldElement fieldElement) {
+    if (jsonKeyNameValue != null) {
+      return jsonKeyNameValue;
+    }
+
+    switch (classAnnotation.fieldRename) {
```
natebosch: here's another possible way to handle this which leaves more room for extension:

change the `fieldRename` argument from an `enum` to a `String Function(String)`. Change the top-level method `snakeCase` to a static method `FieldRename.snake`.

- Usage looks identical: `@JsonSerializable(fieldRename: FieldRename.snake)`
- No need for a `switch` or needing to touch lots of points of code to add new options.
- Users can add their own rename logic by implementing a `String Function(String)`.

kevmoo: @natebosch , you're crazy – like a 🦊 

Would likely help headed toward https://github.com/dart-lang/json_serializable/issues/280 , too

kevmoo: Well...I'll try. The function is defined in user code, but needs to be invoked in the context of the build action – so...maybe we can make it work with mirrors?

We'll see...

natebosch: hmm, yeah that might not be feasible - it puts us in really weird places when it comes to invalidation of the build script - we might not invalidate the outputs if the users changes the implementation of the rename function.

We could only do this, I think, if the rename were lazy and done at serialization time - which also means we'd need a pair of functions to go in both directions...

PR #283, opened 2018-08-13
misc: throw more specific error for some classes of bad ctor

kevmoo:
Also migrated a number of tests to annotation-based description
   ...much easier to maintain

### DIFF
--- a/json_serializable/test/json_serializable_test.dart
+++ b/json_serializable/test/json_serializable_test.dart
@@ -13,6 +13,16 @@ import 'package:test/test.dart';
 import 'analysis_utils.dart';
 import 'test_file_utils.dart';
 
+Matcher _matcherFromShouldGenerateAnnotation(ConstantReader reader) {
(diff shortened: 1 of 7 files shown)


json_serializable/test/json_serializable_test.dart
```diff
@@ -13,6 +13,16 @@ import 'package:test/test.dart';
 import 'analysis_utils.dart';
 import 'test_file_utils.dart';
 
+Matcher _matcherFromShouldGenerateAnnotation(ConstantReader reader) {
+  var expectedOutput = reader.read('expectedOutput').stringValue;
+  var isContains = reader.read('contains').boolValue;
+
+  if (isContains) {
```
natebosch: [optional] consider using ternary conditional since the interesting bit is _what_ it is, not the side effects.

kevmoo: Will do in follow-up

PR #293, opened 2018-08-24
 misc: document behavior of DecodeHelper.createFactory

kevmoo:


kevmoo:
Nevermind – doing something better here...

PR #302, opened 2018-09-01
fix: cast to Set if target type is Set

gagoman:
Fixes https://github.com/dart-lang/json_serializable/issues/54

### DIFF
--- a/json_serializable/test/src/_json_serializable_test_input.dart
+++ b/json_serializable/test/src/_json_serializable_test_input.dart
@@ -120,6 +120,21 @@ class FinalFieldsNotSetInCtor {
   FinalFieldsNotSetInCtor();
 }
 
+@ShouldGenerate(r'''
+SetSupport _$SetSupportFromJson(Map<String, dynamic> json) {
+  return SetSupport((json['values'] as List)?.map((e) => e as int)?.toSet());
+}
+
+Map<String, dynamic> _$SetSupportToJson(SetSupport instance) =>
+    <String, dynamic>{'values': instance.values?.toList()};
+''')
+@JsonSerializable()
+class SetSupport {
+  final Set<int> values;
+
+  SetSupport(this.values);
+}
+
 @JsonSerializable(createToJson: false)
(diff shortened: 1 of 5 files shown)


json_serializable/test/src/_json_serializable_test_input.dart
```diff
@@ -120,6 +120,21 @@ class FinalFieldsNotSetInCtor {
   FinalFieldsNotSetInCtor();
 }
 
+@ShouldGenerate(r'''
+SetSupport _$SetSupportFromJson(Map<String, dynamic> json) {
+  return SetSupport((json['values'] as List)?.map((e) => e as int)?.toSet());
```
gagoman: I was thinking to modify first cast (`as List`) to `as Iterable` but realized that `dart:convert` returns list anyway

json_serializable/test/src/_json_serializable_test_input.dart
```diff
@@ -120,6 +120,21 @@ class FinalFieldsNotSetInCtor {
...
+}
+
+Map<String, dynamic> _$SetSupportToJson(SetSupport instance) =>
+    <String, dynamic>{'values': instance.values?.toList()};
```
gagoman: I have not changed return type cast (from `toList` to `toSet`) as `Set` is not supported by default by `dart:convert`

gagoman:
Any other tests I should add?

json_serializable/test/src/_json_serializable_test_input.dart
kevmoo: yup

json_serializable/test/src/_json_serializable_test_input.dart
kevmoo: yup

gagoman:
Done

gagoman:
What are the cases? I can work on them.

kevmoo:
> What are the cases? I can work on them.

https://github.com/dart-lang/json_serializable/pull/303/files#r215043811

PR #303, opened 2018-09-04
Follow-up to set-support work

kevmoo:
Related to https://github.com/dart-lang/json_serializable/issues/54
Follow-up on https://github.com/dart-lang/json_serializable/pull/302


json_serializable/lib/src/type_helpers/iterable_helper.dart
```diff
@@ -70,8 +70,10 @@ class IterableHelper extends TypeHelper {
 
     var itemSubVal = context.deserialize(iterableGenericType, closureArg);
 
-    // If `itemSubVal` is the same, then we don't need to do anything fancy
-    if (closureArg == itemSubVal) {
+    // If `itemSubVal` is the same and it's not a Set, then we don't need to do
+    // anything fancy
+    if (closureArg == itemSubVal &&
```
kevmoo: @gagoman – this was the one case that was missing. Tough to find unless you do the full set of cases in the `kitchen_sink` test

json_serializable/test/kitchen_sink/kitchen_sink.dart
```diff
@@ -12,6 +12,7 @@ import 'strict_keys_object.dart';
 part 'kitchen_sink.g.dart';
 
 List<T> _defaultList<T>() => null;
```
jakemac53: Can we get a comment here that these get overwritten to not return null in some cases :D

jakemac53: Alternatively could you use like a zone variable or something to make this a bit less hacky?

kevmoo: I'll add a comment. Let's keep the runtime clean.

PR #307, opened 2018-09-10
Export all TypeHelper implementations in type_helper.dart

eredo:
Export all helpers in type_helper.dart so they can be reused in other packages (like dartson).
See https://github.com/eredo/dartson/issues/48#issuecomment-419239742 for original discussion.


eredo:
@kevmoo done. Wasn't sure regarding the correct version.

PR #318, opened 2018-09-19
Replace *erializeContext with TypeHelperContext

kevmoo:
Eliminates special knowledge of internal implementation types from
exported TypeHelpers

### DIFF
--- a/json_serializable/lib/src/type_helper.dart
+++ b/json_serializable/lib/src/type_helper.dart
@@ -5,25 +5,34 @@
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 
-abstract class SerializeContext {
-  /// `true` if [serialize] should handle the case of `expression` being null.
-  bool get nullable;
+/// Context information provided in calls to [TypeHelper.serialize] and
+/// [TypeHelper.deserialize].
+abstract class TypeHelperContext {
+  /// The annotated class that code is being generated for.
+  ClassElement get classElement;
+
+  /// The field that code is being generated for.
+  FieldElement get fieldElement;
+
+  /// Represents the corresponding configuration on `JsonSerializableGenerator`.
+  bool get explicitToJson;
+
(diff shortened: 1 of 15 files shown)


kevmoo:
CC @eredo

json_serializable/lib/src/type_helper.dart
```diff
@@ -5,25 +5,34 @@
...
-  bool get nullable;
+/// Context information provided in calls to [TypeHelper.serialize] and
+/// [TypeHelper.deserialize].
+abstract class TypeHelperContext {
+  /// The annotated class that code is being generated for.
+  ClassElement get classElement;
+
+  /// The field that code is being generated for.
+  FieldElement get fieldElement;
+
```
eredo: How about providing a `dynamic get parentContext` instead of the following getters (`explicitToJson`, `anyMap`, `useWrappers`) and do an instanceof-check (in this case `JsonSerializableGenerator`) within the `TypeHelper` implementation to fetch the properties. This way further configurations of `TypeHelperContext` outside of `json_serializable` could be passed to their own `TypeHelper` implementations using this context. While writing this comment, I realize that this instanceof-check could be done on the hole `TypeHelperContext` implementation, but still these 3 getters would than be obsolete.

kevmoo: I'm going to look at that next. I hear you. I really don't like `dynamic` in this case, though.

I'd rather have a generic parameter (or something) because folks invoking `TypeHelper` instances with these setting should be able to provide them...

kevmoo: See what I did here: https://github.com/dart-lang/json_serializable/commit/8b9cdb94410dce2f27c7d759ea138ed4dc7906f4#diff-149ae6bf88148302b5d60654f7f7c36cR17

I want to avoid auto-magic casting because then other generators don't have any way to plug in.

kevmoo: @eredo see https://github.com/dart-lang/json_serializable/pull/319

PR #319, opened 2018-09-19
Update TypeHelper to allow implementations to require richer context

kevmoo:
Replaces approach taken in 8b9cdb94410dce2f27c7d759ea138ed4dc7906f4
Much easier for implementations to run with other generators


kevmoo:
CC @eredo

eredo:
Really like this!

PR #329, opened 2018-09-29
Use final when possible.

lejard-h:
see https://github.com/dart-lang/json_serializable/issues/290

### DIFF
--- a/example/lib/example.g.dart
+++ b/example/lib/example.g.dart
@@ -19,7 +19,7 @@ Person _$PersonFromJson(Map<String, dynamic> json) {
 }
 
 Map<String, dynamic> _$PersonToJson(Person instance) {
-  var val = <String, dynamic>{
+  final val = <String, dynamic>{
     'firstName': instance.firstName,
   };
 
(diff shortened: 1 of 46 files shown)


kevmoo:
Hrm. I'm sympathetic to this, but I worry we are straying from our general coding style.

Thoughts, @jakemac53 @natebosch ?

kevmoo:
@lejard-h – also, some analyzer cleanup needed - https://travis-ci.org/dart-lang/json_serializable/jobs/434942456#L567

jakemac53:
I don't have a strong opinion here - as long as there is a lint for it to require it doesn't regress which it looks like is the case. At least it enforces consistency... and it is technically a bit safer and more self documenting.

lejard-h:
How am I suppose to pass the ensure_build_test inside example ?
it does not seem to use the modificated version of json_serializable builder

kevmoo:
> How am I suppose to pass the ensure_build_test inside example ?
it does not seem to use the modificated version of json_serializable builder

It's tough here, since you're updating the lints that affect all projects, but the example doesn't have it's updated generator.

You might need to comment out the lint until we can land this and roll the examples.

example/lib/example.g.dart
```diff
@@ -19,7 +19,7 @@ Person _$PersonFromJson(Map<String, dynamic> json) {
 }
 
 Map<String, dynamic> _$PersonToJson(Person instance) {
-  var val = <String, dynamic>{
+  final val = <String, dynamic>{
```
kevmoo: Just revert all changes to `example` – we'll have to update it later - https://travis-ci.org/dart-lang/json_serializable/jobs/436810474#L551

PR #336, opened 2018-10-03
Support decode/encode of dart:core Duration

lejard-h:
fix https://github.com/dart-lang/json_serializable/issues/310

add `$parseDuration`  method since `Duration.parse` does not exist (https://github.com/dart-lang/sdk/issues/16059)


kevmoo:
Thanks, @lejard-h !

PR #345, opened 2018-10-10
Support all builder configuration via class annotation

kevmoo:
Eliminate the (never-published) GeneratorConfig class
All builder and class configuration is now centralized in JsonSerializable
Also removed the custom `tool/build.dart` file – all configured via build.yaml now


json_annotation/lib/src/json_serializable.g.dart
```diff
@@ -1,26 +1,26 @@
 // GENERATED CODE - DO NOT MODIFY BY HAND
 
-part of 'generator_config.dart';
```
kevmoo: This was copy-pasted as part of the PR – generation is not setup (yet). Going to do that later.

PR #346, opened 2018-10-10
Fix new instances of unnecessary_this in latest SDK

kevmoo:
Fixed a number of prefer_final_locals outside of the example pkg


example/lib/example.dart
```diff
@@ -23,7 +23,7 @@ class Person {
 
   Person(this.firstName, this.lastName, this.dateOfBirth,
       {this.middleName, this.lastOrder, List<Order> orders})
-      : this.orders = orders ?? <Order>[];
+      : orders = orders ?? <Order>[];
```
jakemac53: hmmm I actually much prefer the `this` here for disambiguation purposes

jakemac53: :man_shrugging:

kevmoo: Yeah, this confused the heck out of me, too.

I guess if it's the value being assigned it **has** to be for `this` – but it is a bit weird.

jakemac53: Ya it is technically not necessary but I always have preferred it....

PR #358, opened 2018-12-06
Fix documentation comment typo

jonasfj:


json_annotation/lib/src/json_serializable.dart
```diff
@@ -54,7 +54,7 @@ class JsonSerializable {
   /// ```
   final bool createFactory;
 
-  /// If `true` (the default), code for decoding JSON is generated fon this
+  /// If `true` (the default), code for encoding JSON is generated for this
```
jonasfj: I assming this was meant to say "encoding" instead of "decoding"

PR #360, opened 2018-12-10
Return defaultValue for unsupported enum instead throw exception

beastoin:
For enum type, I think we should use `nullable` option for both `null` and nonexistent value. 

In our case, we have object Coupon which contains CouponType as type of specific Coupon. At the very first time, our CouponType is only `loship_promotion`. We also support unknown CouponType from UI, which showing a message for updating the application. And now, we have `merchant_promotion`  as a new type. Of course, the number of CouponType's values can be increased in the future. So we need to define the parser for unknown type, and I think this is a common case in the real world.

I'm trying to use `nullable` and `defaultValue` options for enum, but my code throws the exception because of unsupported values.

[json_serializable/json_annotation/lib/src/json_key.dart](json_key.dart)

```
 /// When `true`, `null` values are handled gracefully when
  /// serializing the field to JSON and when deserializing `null` and
  /// nonexistent values from a JSON map.
  ///
  /// Setting to `false` eliminates `null` verification in the generated code
  /// for the annotated field, which reduces the code size. Errors may be thrown
  /// at runtime if `null` values are encountered, but the original class should
  /// also implement `null` runtime validation if it's critical.
  ///
  /// The default value, `null`, indicates that the behavior should be
  /// acquired from the [JsonSerializable.nullable] annotation on the
  /// enclosing class.
  final bool nullable;
```

### DIFF
--- a/json_annotation/lib/src/json_key.dart
+++ b/json_annotation/lib/src/json_key.dart
@@ -26,6 +26,20 @@ class JsonKey {
   /// enclosing class.
   final bool nullable;
 
+  /// When `true`, `null` values are handled gracefully when
+  /// serializing the field to JSON and when deserializing `null` and
+  /// nonexistent values from a JSON map.
+  ///
+  /// Setting to `false` eliminates `null` verification in the generated code
+  /// for the annotated field, which reduces the code size. Errors may be thrown
+  /// at runtime if `null` values are encountered, but the original class should
+  /// also implement `null` runtime validation if it's critical.
+  ///
+  /// The default value, `null`, indicates that the behavior should be
+  /// acquired from the [JsonSerializable.unknowable] annotation on the
+  /// enclosing class.
+  final bool unknowable;
+
   /// `true` if the generator should include the this field in the serialized
   /// output, even if the value is `null`.
   ///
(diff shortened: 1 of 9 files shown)


beastoin:
I signed it!

beastoin:
Oops!!! Unit test failed as the same result to `master` branch, please double check :(

kevmoo:
I'm not a fan of the complexity this introduces. Have you tried using a custom [JsonConverter](https://pub.dartlang.org/documentation/json_annotation/latest/json_annotation/JsonConverter-class.html) for this case?

beastoin:
> I'm not a fan of the complexity this introduces. Have you tried using a custom [JsonConverter](https://pub.dartlang.org/documentation/json_annotation/latest/json_annotation/JsonConverter-class.html) for this case?

Of course, my practice can be archived by using JsonConverter (as the way I did). But I think we can make this library better and easier, just by using as the right way of  field `nullable` and `defaultValue`, specific with nullable concept from
[JsonKey](https://pub.dartlang.org/documentation/json_annotation/latest/json_annotation/JsonKey/nullable.html), we should make our code away from exception for nonexistent value too, but not only for null value.

Thanks and please correct me if I'm wrong.

PR #380, opened 2019-02-01
Support the latest version of pkg:analyzer, prepare v2.0.2 for release

kevmoo:


json_serializable/lib/src/utils.dart
```diff
@@ -46,7 +46,19 @@ void throwUnsupported(FieldElement element, String message) =>
...
 FieldRename _fromDartObject(ConstantReader reader) => reader.isNull
     ? null
-    : FieldRename.values[reader.objectValue.getField('index').toIntValue()];
+    : FieldRename.values.singleWhere(
+        (thing) {
+          var name = thing.toString().split('.')[1];
+          var result = reader.objectValue.getField(name);
+          return result != null;
+        },
+        // TODO: remove once pkg:analyzer < 0.35.0 is no longer supported
```
kevmoo: `pkg:analyzer` changed how `DartObject` looks when it represents an enum value. I did this here and above so we don't have to bump the pkg:analyzer dep right away w/ this release.

I did test back to `0.33.3` to make sure this works across the supported version range.

json_serializable/test/json_serializable_test.dart
```diff
@@ -365,6 +365,7 @@ String _runForElementNamedWithGenerator(
 final _annotatedElements = _library.allElements
     .map<AnnotatedElement>((e) {
       for (final md in e.metadata) {
+        md.computeConstantValue();
```
kevmoo: Zero clue why this changed, but I need to call this or the constant reader doesn't work. 🤷‍♂️

jakemac53: Oh... this could actually be an accidental breaking change in build_resolvers. We used to do it for you and maybe we don't any more.

kevmoo: FWIW, this is just test infrastructure code. All of the actual code generation worked fine...

json_serializable/lib/src/utils.dart
kevmoo: I rolled this into one helper in the latest commit

json_serializable/test/json_serializable_test.dart
jakemac53: hmm, I will keep my eye out for similar bug reports but not worry about it for now then

PR #385, opened 2019-02-14
Generate a conversion expression for fromJson param type

kevmoo:
Previously we just did a cast to the raw param type, which was often wrong

See https://stackoverflow.com/q/54654171/39827


kevmoo:
fixed in https://github.com/dart-lang/json_serializable/commit/1dcd8f353cedff7431bcc84c15b70fe696636e48

PR #386, opened 2019-02-14
Fix typo field name between the comments.

zhengxiaoyao0716:
generate_to_json_function => generateToJsonFunction


zhengxiaoyao0716:
I signed it!

zhengxiaoyao0716:
@kevmoo Thanks your suggestion, although I could, but maybe it's not a nice idea, for the other codes not use the `[]`. If you sure we should change it use [] syntax, maybe the other codes should changed too?

image
image
image

kevmoo:
If you want to dive in there, that'd be great!

kevmoo:
Closing out per discussion.

PR #393, opened 2019-02-25
Fix travis for Dart 2.0.0 stable

kevmoo:
Guessing configuration-specific imports
(now used by pkg:package_resolver)
are broken on DDC on Dart 2.0.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,13 @@ jobs:
       env: PKG="json_serializable"
       dart: stable
     - stage: unit_test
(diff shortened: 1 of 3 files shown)


kevmoo:
FYI @natebosch – seems like the config-specific update to `package_resolver` breaks DDC on 2.0.0

Uh...we still might want to do a 1.0.9 that just reverts the change for SDK 2.0.0 and then do what @lrhn says for a 1.0.10 – drop support for SDK 2.0.0

PR #401, opened 2019-03-04
Migrate to pkg:source_gen_test

kevmoo:


json_serializable/test/src/_json_serializable_test_input.dart
```diff
@@ -7,33 +7,42 @@
 import 'dart:collection';
 
 import 'package:json_annotation/json_annotation.dart';
-
-import 'annotation.dart';
+import 'package:source_gen_test/annotations.dart';
 
 part 'checked_test_input.dart';
-part 'core_subclass_type_input.dart';
+
```
natebosch: why include a bunch of blank lines? This breaks certain editor behavior like sorting a block of lines.

kevmoo: this was a weird behavior of the autofix in intellij. will revert

PR #403, opened 2019-03-08
Complex map formatting

kevmoo:


kevmoo:
~20x faster on generated test code

PR #408, opened 2019-03-13
test: move KitchenSink to just use test functions

kevmoo:


kevmoo:
The mixin model is pretty much ignored at this point...

PR #411, opened 2019-03-14
Refactor kitchen sink tests

kevmoo:
Have each test `factory` describe itself
Run the right tests given the factor configuration
Will make it much easier to add tests and configurations later


natebosch:
The commit message should describe what changed.

How is this an improvement?

kevmoo:
> The commit message should describe what changed.
> 
> How is this an improvement?

Updated description. I'm headed in a direction here.

Actually. Hold off reviewing this. I need to fix a couple more things.

kevmoo:
> I didn't look too closely - this seems like a rather big change and it's not super clear to me what the value is, but since it's test code and I don't see anything egregious I'm not going to spend much time on it.

The value all comes later when adding new configurations. In the short-term it's just cleanup.

PR #413, opened 2019-03-14
Add support for dart:core BigInt type

kevmoo:
Fixes https://github.com/dart-lang/json_serializable/issues/412

### DIFF
--- a/json_serializable/CHANGELOG.md
+++ b/json_serializable/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.1.0
+
+* Added support for `BigInt`.
+
+* Added `BigIntTypeHelper` to `type_helper.dart` library.
+
 ## 2.0.3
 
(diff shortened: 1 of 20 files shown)


eseidel:
Looks like I need something like this: https://github.com/lrhn/json-tool/blob/master/example/bigint_json.dart

eseidel:
Nevermind, turns out we already had some fancy for this:
```dart

/// {@template big_int_converter}
/// Converts between [BigInt] and [num]. If a number is representable as an
/// [int], it will be serialized as an [int]. Otherwise, it will be serialized
/// as a [double].
/// {@endtemplate}
class BigIntConverter implements JsonConverter<BigInt, num> {
  /// {@macro big_int_converter}
  const BigIntConverter();

  @override
  BigInt fromJson(num json) {
    return BigInt.from(json);
  }

  @override
  num toJson(BigInt object) {
    return object.toInt();
  }
}
```

PR #443, opened 2019-04-10
Handle the intersection of convert function, !nullable & !includeIfNull

kevmoo:
If a field has a conversion function defined – either `JsonKey.toJson`
or a custom `JsonConverter` annotation – handle the case where the
function returns `null` and both `nullable` and `includeIfNull` are
`false`.

Before this change, you could still get `null` values in output JSON
even if `includeIfNull` is `false`.

Prepare to release v2.2.0


kevmoo:
@jakemac53 – FYI, I'm going to ditch wrapper support. Too darn complicated! https://github.com/dart-lang/json_serializable/issues/441

PR #445, opened 2019-04-10
Removed `JsonSerializable.useWrappers`

kevmoo:


kevmoo:
V3 is ALL about removing underused features that are a bear to maintain – and make it hard to add new features

PR #454, opened 2019-04-18
master

humazed:
Use `.toString()` in parsing `String` instead of `as String` in value_helper.dart

parse to `double` or `int` even if the type is `String` as long as the  `int.parse()` or `int.parse()` succeed


humazed:
I signed it!

PR #460, opened 2019-04-30
yaml_decode: add example and associated tests

kevmoo:


yaml_decode/example/example.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:json_annotation/json_annotation.dart';
```
jakemac53: This example doesn't directly use any apis from this package?

kevmoo: The annotation, my friend. 😉

jakemac53: I get that JsonSerializable uses it, but that doesn't serve as an example for using this package?

kevmoo: Sorry. Doing crawl walk run. Going to add in a CLI here that parses yaml.

kevmoo: ...and associated tests, etc

jakemac53: Is it worth having this at all for now though?

kevmoo: Trying to keep the PRs smaller. I'm going to build on this next – I promise. 😄

PR #461, opened 2019-04-30
yaml_decode: fill out example and add tests

kevmoo:


yaml_decode/test/example_test.dart
```diff
@@ -0,0 +1,123 @@
...
+  return TestProcess.start(
+    dartPath,
+    ['example/example.dart', _testYamlPath],
+  );
+}
+
+final String dartPath = p.join(_sdkDir, 'bin', 'dart');
+
+/// The path to the root directory of the SDK.
+final String _sdkDir = (() {
```
jakemac53: This is overkill for testing code imo (we don't do this extra logic anywhere else), although its fine if you want to leave it. I would just do `final _sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));` though.

kevmoo: Eh...maybe. Sometimes you want to run w/ an SDK not on your path. I get it.

Need to add these utils to pkg:io

PR #466, opened 2019-05-02
Update readme, prepare to release yaml_decode v1

kevmoo:
Fixes https://github.com/dart-lang/json_serializable/issues/437

### DIFF
--- a/yaml_decode/readme.md
+++ b/yaml_decode/readme.md
@@ -1,1 +1,71 @@
 [![Pub Package](https://img.shields.io/pub/v/yaml_decode.svg)](https://pub.dartlang.org/packages/yaml_decode)
+
+`package:yaml_decode` provides a `toParsedYamlException` function to create a
+`ParsedYamlException` that contains source and error information that helps
+provide actionable errors when decoding invalid YAML. 
+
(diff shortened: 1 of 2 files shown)


kevmoo:
@natebosch – you seem the kinda guy who would want to chime in on the package name

Thoughts?

yaml_decode/readme.md
```diff
@@ -1 +1,71 @@
 [![Pub Package](https://img.shields.io/pub/v/yaml_decode.svg)](https://pub.dartlang.org/packages/yaml_decode)
+
+`package:yaml_decode` provides a `toParsedYamlException` function to create a
+`ParsedYamlException` that contains source and error information that helps
```
jakemac53: `ParsedYamlException` from a `CheckedJsonException`

natebosch:
It looks like the reason this exists as a separate package is because we don't want add a dependency from `json_annotation` to `yaml` or vice versa.

Can we consider changing the signature of `toParsedYamlException` such that we pass the individual fields from the `CheckedFromJsonException` instead of that type? Then this code could live in `package:yaml` right?

kevmoo:
Going to rename the package to `checked_yaml` – see the last PR. We have a better plan here.

PR #471, opened 2019-05-02
Readme, test tweaks, prepare to release

kevmoo:


checked_yaml/example/example.dart
```diff
@@ -36,10 +36,18 @@ class Configuration {
 }
 
 void main(List<String> arguments) {
-  final fileContents = File(arguments.single).readAsStringSync();
+  var sourceUrl = arguments.single;
```
jakemac53: nit: it looks like this is either a source url or a yaml string, consider renaming it if so? `sourceUrlOrYaml` or something.

Or, just drop support for one or the other, I think it distracts from the example.

checked_yaml/test/example_test.dart
```diff
@@ -2,171 +2,111 @@
...
-import 'package:path/path.dart' as p;
+import 'package:checked_yaml/checked_yaml.dart';
 import 'package:test/test.dart';
-import 'package:test_descriptor/test_descriptor.dart' as d;
-import 'package:test_process/test_process.dart';
 
-void main() {
-  test('valid input', () async {
-    final proc = await _run('{"name": "bob", "count": 42}');
+import '../example/example.dart' as example;
```
jakemac53: This is pretty naughty, you won't ever be able to use `build_vm_compilers` if you do this, but :man_shrugging:

kevmoo: agreed 🤷‍♂

checked_yaml/example/example.dart
kevmoo: It's kinda nice to support either – at least it makes my tests work.

Will rename.

kevmoo: fixed!

checked_yaml/test/example_test.dart
kevmoo: @jakemac53 to file a bug!

PR #482, opened 2019-05-10
remove encode_empty_collection

kevmoo:
Fixes https://github.com/dart-lang/json_serializable/issues/446


kevmoo:
I know I just added this, but I found a model that's pretty easy for folks to implement themselves and it removes a LOT of magic here

PR #491, opened 2019-05-23
2 cleanup commits

kevmoo:


json_serializable/lib/src/type_helpers/to_from_string.dart
```diff
@@ -0,0 +1,62 @@
...
+const uriString = ToFromString(
+  'Uri.parse',
+  'toString()',
+  TypeChecker.fromUrl('dart:core#Uri'),
+);
+
+/// Helper that unifies implementations of [Type]s that convert trivially
+/// to-from [String].
+class ToFromString {
+  final String _parse;
```
jakemac53: I would add some more docs here, these are string representations of the code to convert to/from strings yes?

json_serializable/lib/src/type_helpers/to_from_string.dart
```diff
@@ -0,0 +1,62 @@
...
+
+const uriString = ToFromString(
+  'Uri.parse',
+  'toString()',
+  TypeChecker.fromUrl('dart:core#Uri'),
+);
+
+/// Helper that unifies implementations of [Type]s that convert trivially
+/// to-from [String].
+class ToFromString {
```
jakemac53: This name was really confusing to me at first, although I can't come up with a better one off the top of my head... maybe `StringSerializer` but that isn't really accurate either.

kevmoo: Made it `ToFromStringHelper` 🤷‍♂ – this is very package-internal

json_serializable/lib/src/type_helpers/to_from_string.dart
kevmoo: done.

PR #494, opened 2019-05-24
Add trailing commas to constructor invocations and Map literals

kevmoo:
Subjective, but it seems to make the generated code more readable

### DIFF
--- a/json_serializable/test/src/_json_serializable_test_input.dart
+++ b/json_serializable/test/src/_json_serializable_test_input.dart
@@ -59,8 +59,7 @@ class OnlyStaticMembers {
   static int get understand => 42;
 }
 
-@ShouldGenerate(
-  r'''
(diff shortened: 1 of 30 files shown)


json_serializable/test/src/_json_serializable_test_input.dart
```diff
@@ -456,11 +454,9 @@ Map<String, dynamic> _$PropInMixinI448RegressionToJson(
         PropInMixinI448Regression instance) =>
     <String, dynamic>{
       'nullable': instance.nullable,
-      'notNullable': instance.notNullable
+      'notNullable': instance.notNullable,
     };
-''',
-  configurations: ['default'],
```
kevmoo: This is just cleanup from previous work – not needed!

kevmoo:
PTAL

PR #515, opened 2019-07-30
Enum convert map: Remove explicit type and add trailing comma

kevmoo:
Also added a mismatched value type to json_test_common


json_serializable/test/integration/json_test_example.g.dart
```diff
@@ -67,14 +67,14 @@ T _$enumDecodeNullable<T>(Map<T, dynamic> enumValues, dynamic source) {
   return _$enumDecode<T>(enumValues, source);
 }
 
-const _$CategoryEnumMap = <Category, dynamic>{
+const _$CategoryEnumMap = {
```
kevmoo: Inference will give us a tighter type.

PR #519, opened 2019-08-01
Add explicit type arguments to generic instantiations

kevmoo:
- Enabled the `strict-raw-types` analysis option.
- Update the generator to generate compliant code
- Update all hand-written code

Fixes https://github.com/dart-lang/json_serializable/issues/517


kevmoo:
@natebosch if you want to run with this, I got you started - https://github.com/dart-lang/json_serializable/issues/517

kevmoo:
CC @srawlins

example/lib/example.g.dart
```diff
@@ -15,7 +15,7 @@ Person _$PersonFromJson(Map<String, dynamic> json) {
     lastOrder: json['last-order'] == null
         ? null
         : DateTime.parse(json['last-order'] as String),
-    orders: (json['orders'] as List)
+    orders: (json['orders'] as List<dynamic>)
```
srawlins: FYI, we are skipping the `is` and `as` checks on strict-raw-types internally at Google. 99% of the time, you're looking at unstructured parsed data, like this JSON, and your only choice is to write `as List<dynamic>`, and you've neither improved readability nor type promotion. :/ In analysis_options.yaml you'd ignore `STRICT_RAW_TYPE_IN_AS` and `STRICT_RAW_TYPE_IN_IS`.

natebosch: If we are not using them internally and recommending against them externally - should they be dropped altogether?

kevmoo:
> I consider this a total failure of strict-raw-types, when literally every single implicit dynamic is simply replaced with an explicit dynamic. strict-raw-types is here to uncover casting bugs, like `List a = [1, 2, 3]`, but that did not happen anywhere in this CL. :(
> 
> Maybe this is just expected for a JSON package, but not a shining moment for strict-raw-types.

If this PR is dropped and just as fodder to reconsider, I still consider it a win.

I cringed through this thinking the same things...

example/lib/example.g.dart
srawlins: We're not using them internally because it looks like they add more noise than benefit, and thats a lot to _push_ onto developers. But I can see a dev finding these useful on their own individual pub package, and wanting the checks.

I'm not recommending against them externally; I'm just mentioning what we do internally, in case it is useful information.

kevmoo:
CC @leafpetersen

srawlins:
I could see extension methods helping with JSON APIs, and generalized typedefs doing cool things to turn the volume down on strict-raw-types:

```dart
typedef JsonMap = Map<dynamic, dynamic>;
extension Something on JsonMap {
  List<T> getList<T>(dynamic key) => (this[key] as List).cast<T>();
  // ...
}

// Or
@optionalTypeArgs
typedef JsonList<T>=  List<T>;
extension Something on JsonList { ... }
```

You could then even enable STRICT_RAW_TYPE_IN_IS because it would encourage the developer to use JsonList and JsonMap, etc in code, which I like as a human. Often I  look at code that I think deals with JSON,  and I see a Map being passed around, but is that like, a Map from  the JSON  stuff, or another Map as part of the API? Seeing JsonMap is like, boom. I know what should be inside  there, and where it came from.

natebosch:
Once I filter out `strict_raw_types_in_as` and `strict_raw_types_in_is` it looks like the only remaining fix we need in this package is to change `fromJson(Map json)` to `fromJson(Map<dynamic, dynamic> json)` in generate constructors and utility methods. This might apply only in the case of `anyMap: true`.

example/lib/example.g.dart
leafpetersen: > I'm just mentioning what we do internally, in case it is useful information.

Honestly, that seems like pretty good data that they're not worth their weight.

mnordine:
> Once I filter out `strict_raw_types_in_as` and `strict_raw_types_in_is`

How do you do that?

srawlins:
You can edit your analysis_options.yaml file: https://dart.dev/guides/language/analysis-options

mnordine:
> You can edit your analysis_options.yaml file: https://dart.dev/guides/language/analysis-options

Tried that with `strict_raw_types_in_as` and `strict_raw_types_in_is` with latest dev release.  
Analyzer doesn't understand those fields  

(I already have `strict-raw-types`, obviously)

srawlins:
What did you try? I was specifically referring to setting them to 'ignore': https://dart.dev/guides/language/analysis-options#ignoring-rules

mnordine:
```yaml
  errors:
    strict_raw_types_in_as: ignore
    strict_raw_types_in_is: ignore
```

and other variants

srawlins:
Sorry, it's singular: STRICT_RAW_TYPE_IN_AS

mnordine:
That works, thx

mnordine:
No longer works with `2.6.0-dev.2.0`

srawlins:
Correct the STRICT_RAW_TYPE_IN_AS and STRICT_RAW_TYPE_IN_IS Hints were removed.

mnordine:
Why? They seemed helpful in some circumstances, as noted above. Was it because `strict-raw-types` ignores them by default now? 😕

srawlins:
> Why?

See the discussion above, https://github.com/dart-lang/json_serializable/pull/519/files/1b933420c81d7caae33794211bd8bf532731eb54#diff-a8a492e404f47e94b56e1f302b7b1a51.

> They seemed helpful in some circumstances, as noted above.

I cannot find the comment(s) you are referring to.

> Was it because strict-raw-types ignores them by default now? 😕

Yes.

mnordine:
I think there's a bit of a miscommunication. I was referring to the discussion you linked to. It seems we're in agreement, and not having those warnings with `strict-raw-types` is the right choice, 👍

kevmoo:
I'm closing this out, folks!

PR #534, opened 2019-08-21
support latest pkg:analyzer, prepare to release 3.2.1

kevmoo:
Note: doing a cast because `enclosingElement` is now `Element` as of
pkg:analyzer v0.38.0

### DIFF
--- a/json_serializable/lib/src/field_helpers.dart
+++ b/json_serializable/lib/src/field_helpers.dart
@@ -37,7 +37,10 @@ class _FieldSet implements Comparable<_FieldSet> {
   int compareTo(_FieldSet other) => _sortByLocation(sortField, other.sortField);
 
   static int _sortByLocation(FieldElement a, FieldElement b) {
-    final checkerA = TypeChecker.fromStatic(a.enclosingElement.type);
+    final checkerA = TypeChecker.fromStatic(
+        // TODO: remove `ignore` when min pkg:analyzer >= 0.38.0
+        // ignore: unnecessary_cast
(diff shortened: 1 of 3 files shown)


json_serializable/lib/src/field_helpers.dart
```diff
@@ -37,7 +37,10 @@ class _FieldSet implements Comparable<_FieldSet> {
   int compareTo(_FieldSet other) => _sortByLocation(sortField, other.sortField);
 
   static int _sortByLocation(FieldElement a, FieldElement b) {
-    final checkerA = TypeChecker.fromStatic(a.enclosingElement.type);
+    final checkerA = TypeChecker.fromStatic(
+        // TODO: remove `ignore` when min pkg:analyzer >= 0.38.0
+        // ignore: unnecessary_cast
+        (a.enclosingElement as ClassElement).type);
```
grouma: I was told this will cause issues with extension methods. Are we sure support will come in a breaking change?

kevmoo: As long as `type` remains a member on `ClasssElement` we're all good, right?

natebosch: Not if `a.enclosingElement` starts returning an `ExtensionElement` instead of a `ClassElement`. In that case the cast will fail.

kevmoo: Ah! I'm not worried. This is for concrete fields, not fields added via extension.

PR #535, opened 2019-08-22
Support JsonConverter annotations on property getters

tvolkert:
This fixes a bug whereby the following code:

```dart
@JsonSerializable(createFactory: false)
class JsonConverterOnGetter {
  @JsonKey()
  @NeedsConversionConverter()
  NeedsConversion get annotatedGetter => NeedsConversion();
}
```

would yield the following error:

```
  Could not generate `toJson` code for `annotatedGetter`.
  None of the provided `TypeHelper` instances support the defined type.
  package:<file>:<line>:<column>
      ╷
  ... │   NeedsConversion get annotatedGetter => NeedsConversion();
      │                       ^^^^^^^^^^^^^^^
      ╵
```

### DIFF
--- a/json_serializable/lib/src/type_helpers/json_converter_helper.dart
+++ b/json_serializable/lib/src/type_helpers/json_converter_helper.dart
@@ -76,7 +76,11 @@ _JsonConvertData _typeConverter(DartType targetType, TypeHelperContext ctx) {
   var matchingAnnotations = converterMatches(ctx.fieldElement.metadata);
 
   if (matchingAnnotations.isEmpty) {
-    matchingAnnotations = converterMatches(ctx.classElement.metadata);
+    matchingAnnotations = converterMatches(ctx.fieldElement.getter?.metadata ?? []);
+
+    if (matchingAnnotations.isEmpty) {
+      matchingAnnotations = converterMatches(ctx.classElement.metadata);
+    }
   }
 
   return _typeConverterFrom(matchingAnnotations, targetType);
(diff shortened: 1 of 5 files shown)


tvolkert:
/cc @natebosch @kevmoo 

FYI @dnfield this is the bug that was preventing me from writing a `Key` converter several weeks back.

kevmoo:
one more format @tvolkert

tvolkert:
Oops - done!

json_serializable/test/src/json_converter_test_input.dart
```diff
@@ -164,3 +164,30 @@ class _ConverterWithCtorParams implements JsonConverter<Duration, int> {
   @override
   int toJson(Duration object) => 0;
 }
+
+@ShouldGenerate(r'''
```
kevmoo: Thanks for adding the regression test!

json_serializable/lib/src/type_helpers/json_converter_helper.dart
```diff
@@ -76,7 +76,12 @@ _JsonConvertData _typeConverter(DartType targetType, TypeHelperContext ctx) {
   var matchingAnnotations = converterMatches(ctx.fieldElement.metadata);
 
   if (matchingAnnotations.isEmpty) {
-    matchingAnnotations = converterMatches(ctx.classElement.metadata);
+    matchingAnnotations =
+        converterMatches(ctx.fieldElement.getter?.metadata ?? []);
+
+    if (matchingAnnotations.isEmpty) {
```
natebosch: I might have avoided the nesting, no need to fix if travis is happy but if we need to push another commit we may as well unnest these.

kevmoo: I agree to this from a code style point of view, although I am pondering if it'd have any runtime effect.

tvolkert: Looks like Travis is green.  Happy to push another commit if you'd like?

kevmoo:
Sure! Might s well...

json_serializable/lib/src/type_helpers/json_converter_helper.dart
natebosch: probably a slight runtime effect, but indentation scares me 😝 

this is fine as is.

PR #550, opened 2019-10-17
Don't use DartType.toString() to build Dart code for types.

scheglov:
We would like to change `DartType.toString()` to include nullability suffixes.
It has never been a really good idea to use `toString()` as a way to generate valid Dart code.
So, I'm replacing it with a custom-made `typeToCode` in this package.

This PR has enough changes to pass all tests with these future `analyzer` changes.

### DIFF
--- a/json_serializable/lib/src/helper_core.dart
+++ b/json_serializable/lib/src/helper_core.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/element/element.dart';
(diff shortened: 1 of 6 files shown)


kevmoo:
@scheglov – do we need to change the supported pkg:analyzer range for this?

scheglov:
@kevmoo No. This change is backward compatible, I tested it with the version of `analyzer` that the package depends now, and with dependency overrides to point it to the `analyzer` with my changes to `DartType.toString()`.

PR #553, opened 2019-10-25
Refactor return of boolean literals

natebosch:
For simple conditions without side effects `return condition` is more
straightforward than `if (condition) {return true;} return false;`


kevmoo:
🤷‍♂ this is more clever, but I find it less readable.

Zero effect on runtime, right?

natebosch:
Conventional wisdom is that the refactored version is more readable.

http://cs111.wellesley.edu/~cs111/archive/cs111_spring00/public_html/lectures/boolean-simplification.html
https://medium.com/@ntgard/returning-boolean-literals-is-a-code-smell-7a39531d6b60
https://clang.llvm.org/extra/clang-tidy/checks/readability-simplify-boolean-expr.html

The `&&` or `||` are more readable than trying to work out the "and" or "or" myself based on nesting structure. It makes the value more declarative.

Issue #552, opened 2019-10-25
Upgrade to analyzer 0.39.x


natebosch:
The `InheritanceManager` class is removed in favor of `InheritanceManager3`. I haven't yet figure out the replacement API for `getMembersInheritedFromClasses` is...

Likely related: https://github.com/dart-lang/json_serializable/issues/531

kevmoo:
@scheglov – any ideas?

scheglov:
`getInheritedMap` maybe?

natebosch:
I think that's the likely replacement.

This will take significant refactoring because I need to get a `TypeSystem` passed through to all the places that need it now that didn't before. I also don't think I can synchronously get a `TypeSystem` so everything will need to become async.

scheglov:
If you have a `ResolvedUnitResult`, it has `typeSystem`.
If you have an `Element`, you can ask `element.context.typeSystem`.

natebosch:
> If you have an `Element`, you can ask `element.session.typeSystem`.

This is what I'm doing. Since its async it means any place I do this must be async. I'm trying  a refactor now where I get it once and then hold on to it so the rest of the methods don't need to become async, only the outermost one.

natebosch:
One nasty bit here.

There is a public `TypeSystem` at [`package:analyzer/dart/element/type_system.dart`](https://github.com/dart-lang/sdk/blob/1a611e1d72cccdda70fa32c29fe15de463423cae/pkg/analyzer/lib/dart/element/type_system.dart#L11) however I can't use it.

Classes like [`InheritanceManager3`](https://github.com/dart-lang/sdk/blob/1a611e1d72cccdda70fa32c29fe15de463423cae/pkg/analyzer/lib/src/dart/element/inheritance_manager3.dart#L35) (which the changelog tells me to use over `InheritanceManager`) takes a private [`TypeSystem`](https://github.com/dart-lang/sdk/blob/1a611e1d72cccdda70fa32c29fe15de463423cae/pkg/analyzer/lib/src/generated/type_system.dart#L2375) instead of the public one.

I'll use the `src` import everywhere for now. Is there a plan for a _public_ `InheritanceManager` that I can use without `src` imports?

scheglov:
I'm not aware about any plans to expose `ImheritanceManager` as a public API.
@stereotype441 for future planning.

natebosch:
Unfortunately I think we are going to need a breaking change.

[...]

PR #554, opened 2019-10-25
Upgrade to analyzer 0.39.x

natebosch:
Towards #552

Behavior related to type checks have been moved to `TypeSystem`. Keep an
instance of the type system available on `HelperCore` so that we can
access it synchronously where it is needed.

Remove an `assert` that can no longer be supported due to not having a
`TypeSystem` argument available. It would be breaking to change the
signature to pass one in.

Remove one conditional on equivalent types that can no longer be
performed without a breaking change. This likely also breaks behavior
and we will need to find a solution.

### DIFF
--- a/json_serializable/CHANGELOG.md
+++ b/json_serializable/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.2.4-dev
+
+* Support `package:analyzer` version `0.39.x`.
+
 ## 3.2.3
 
 * Bug fix for analyzer 0.38.5.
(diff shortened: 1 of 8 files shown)


kevmoo:
@natebosch – looks like source_gen needs a bump

kevmoo:
@natebosch – there are build packages that need upgrading, too!

natebosch:
Closing since it's blocked by `source_gen_test` and by https://github.com/dart-lang/json_serializable/issues/552#issuecomment-546542053

We've got all `build` repo packages and `source_gen` updated and published. I've removed the dependency to `json_serializable` from the `build` repo so we aren't blocked by this anymore. We will be blocked if we need to rerun the codegen - as is we're sticking with the currently checked in versions of the files.

kevmoo:
I'll get on source_gen_test!

PR #578, opened 2019-12-21
Enable and fix use_function_type_syntax_for_parameters lint

kevmoo:
Prepare to release +1 releases for serializable and annotation
Bump SDK constraint on annotation

### DIFF
--- a/json_annotation/lib/src/checked_helpers.dart
+++ b/json_annotation/lib/src/checked_helpers.dart
@@ -37,7 +37,7 @@ T $checkedNew<T>(String className, Map map, T Function() constructor,
 /// `JsonSerializableGenerator.checked` is `true`.
 ///
 /// Should not be used directly.
-T $checkedConvert<T>(Map map, String key, T Function(Object value) castFunc) {
+T $checkedConvert<T>(Map map, String key, T Function(Object) castFunc) {
   try {
     return castFunc(map[key]);
   } on CheckedFromJsonException {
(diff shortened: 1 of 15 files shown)


json_annotation/lib/src/checked_helpers.dart
```diff
@@ -37,7 +37,7 @@ T $checkedNew<T>(String className, Map map, T constructor(),
 /// `JsonSerializableGenerator.checked` is `true`.
 ///
 /// Should not be used directly.
-T $checkedConvert<T>(Map map, String key, T castFunc(Object value)) {
+T $checkedConvert<T>(Map map, String key, T Function(Object value) castFunc) {
```
natebosch: [nit] Consider dropping the argument name. `value` is a noise word that was necessary with the old syntax, but the new syntax does not require it.

json_serializable/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.2.4+1
+
+- Fix lint affecting `pub.dev` score.
```
natebosch: This should be `3.2.5`.

json_serializable/lib/src/type_helper_ctx.dart
```diff
@@ -62,7 +62,7 @@ class TypeHelperCtx
       (TypeHelper th) => th.deserialize(targetType, expression, this));
 
   Object _run(DartType targetType, String expression,
-          Object invoke(TypeHelper instance)) =>
+          Object Function(TypeHelper instance) invoke) =>
```
natebosch: [nit] Drop `instance`

json_annotation/lib/src/checked_helpers.dart
kevmoo: ack

json_serializable/lib/src/type_helper_ctx.dart
kevmoo: ack

PR #579, opened 2019-12-23
Method to allow some de-serialisation checks to be overriden at runtime. 

bsutton:
The aim of this pull request is to provide a mechanism to allow a user to override a number of the deserialisation checks at runtime.
The technique used should allow the mechanism to extended to overriding other settings as need arises.

The mechanism allows overrides to be applied globally or on an individual class basis.

The changes are 100% backwards compatible.

I have updated all existing unit tests to work with the changes as well as introducing additional unit tests to ensure the mechanism works as published.

I've provided documentation in the json_serialization projects README.md file on how to use the new mechanism.

The mechanism is easy to use and can be ignored by existing users.

Motivation:
The motivation behind these changes was driven by a desire to use the 'allowedKeys/requiredKkeys' mechanism during development but to relax these settings during production.

I'm happy to make any changes to meet the projects standards.

Brett


.gitignore
```diff
@@ -3,3 +3,4 @@
 .pub
 packages
 pubspec.lock
+.vscode/launch.json
```
kevmoo: Please remove this line. It should be in `~/.gitignore_global`

bsutton: Amended as requested.

example/test/readme_test.dart
```diff
@@ -15,14 +15,14 @@ void _expect(String fileName) {
   test(fileName, () {
     final file = File(fileName);
     expect(file.readAsStringSync(), contains(_pubspecContent));
-  });
+  }, skip: true);
```
kevmoo: You have to copy the contents of the doc/...whatever file into the readme. This makes sure all of the options are linked!

kevmoo: Did you rerun the build?

kevmoo: Oh wait, sorry. Hrm. Let me dig in here. Sorry.

bsutton:
Yes it made no sense to me.

bsutton:
How?

kevmoo:
Run `pub run build_runner build` in `checked_yaml` directory

kevmoo:
```diff
diff --git a/checked_yaml/example/example.g.dart b/checked_yaml/example/example.g.dart
index 854bcba..c1d5447 100644
--- a/checked_yaml/example/example.g.dart
+++ b/checked_yaml/example/example.g.dart
@@ -6,11 +6,12 @@ part of 'example.dart';
 // JsonSerializableGenerator
 // **************************************************************************

-Configuration _$ConfigurationFromJson(Map json) {
+Configuration _$ConfigurationFromJson(Map json, {JsonOverrides overrides}) {
   return $checkedNew('Configuration', json, () {
     $checkKeys(json,
         allowedKeys: const ['name', 'count'],
-        requiredKeys: const ['name', 'count']);
+        requiredKeys: const ['name', 'count'],
+        overrides: overrides);
     final val = Configuration(
       name: $checkedConvert(json, 'name', (v) => v as String),
       count: $checkedConvert(json, 'count', (v) => v as int),
```

kevmoo:
looks like this needs to be updated it `checked_yaml`

example/test/readme_test.dart
bsutton: I've now run the rebuild and commited the generated files.
I had to go through and rationalise the job sdk versions. I've changed them all to stable.
We had a mix of 2.6 and dev which I don't think is necessary now that 2.7 is out.

Everything now appears to be passing except for the readme_test but I can't see how that ever worked as noted previously,

bsutton: worked out the readme problem. the example.g.dart contents needed to be copied into the readme. The test is not an 'equals' but a 'contains' so it does actually work as published.

bsutton:
@kevmoo sorry but I'm not certain what your asking for here!

kevmoo:
I see where you're going with this. I don't like the noise it adds to every generated `fromJson` member, though.

Wondering if this should be enabled in `JsonSerializable` annotation/config

@natebosch, @jakemac53 – thoughts on this? I see where @bsutton is going with it, but I fear the complexity it adds...

json_annotation/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 3.0.2
```
natebosch: If we decide to go forward with this feature it would need to be version `3.1.0`

bsutton:
@natbosh 
 >  this ability to globally override the checks makes me very nervous.

I hadn't consider the implications of globally overriding other packages within the same app.
I would agree that this is probably a bad idea. As such I would be inclined to remove the global override feature. It was a 'nice to have' but given the noted dangers it sounds less nice.

@kevmoo 
> Wondering if this should be enabled in JsonSerializable annotation/config

Are you suggesting this should be an annotation?

I think I've misunderstood you, but if I haven't, then this would defeat the purpose of the changes, which is to allow the validation rules to change at run time.
If I have misunderstood you can you elaborate on your intent.

@natebosch 
> JsonOverrides is too general 

This is a reasonable statement.
ValidationOverrides might be a better choice or just Validation ( I don't like this so much) or just Overrides (this would allow for non-valiation overrides to be added later on).

@natebosch 
One down side to a 'static validation' method is that it then becomes 'global' for a specific serialization class. With the current api you could in theory pick and choose which validation to apply on each call to serialization.
My use case doesn't require it and off hand I can't think of a use case that does except for unit testing.
I"m just a little wary of narrowing down the usage patterns we support without due cause.

@natebosch 
> If we decide to go forward with this feature it would need to be version 3.1.0
Noted.

So if I can sum up the remaining concerns. Essentially there are concerns around potentially polluting the api and should we consider a more general way of controlling validation at run time.

I would suggest that the 'toJson' method will never have validation applied to it.
As such it makes sense to associate any validation with the fromJson method.
If we use the parameter name of 'overrides' then at some point in the future we could add the same parameter to 'toJson' to override non-validation related actions.
This seems like a nice idea as it would make the api symmetrical going forward. 

@natebosch  's idea of static validation isn't terrible but it would be nicer if we could allow a user to control the validation on a per call basis to facilitate unit testing.

I'm don't have the same level of concern regarding 'noise' on the fromJson method. In theory if json_serializable is doing its job then a user of the api never has a real need to even look at the fromJson code and when they are making calls to fromJson they can completely ignore the additional parameter.

So where to from here?

I'm happy to make the suggested changes to the parameter name, version no. and removal of the global override, but I guess we need to reach consensus on whether this PR will get accepted and which of the above discussed validation methods should be used (static or parameter).

natebosch:
> One down side to a 'static validation' method is that it then becomes 'global' for a specific serialization class. With the current api you could in theory pick and choose which validation to apply on each call to serialization.

It would be possible to hack something with Zone variables or similar. If we think that those situations are likely to be edge cases it might be better to let the edge cases look ugly rather than add unused noise in the typical case.

> So if I can sum up the remaining concerns. Essentially there are concerns around potentially polluting the api and should we consider a more general way of controlling validation at run time.

Yeah the main concern for me is the API noise, especially since our hands are tied once we release it. I'm also wary of some of the coupling here, calling out each type of validation by name like this will add extra hoops to jump through to add or change how validation works in the future. For instance the entire approach to null values may need to be updated when we have a Dart version that understands non-nullable types.

> So where to from here?

We might need to wait a bit until @kevmoo can take another look and either hack on a prototype of an alternative or start another discussion about the approach.

kevmoo:
Closing this out – introduces far too much complexity.

PR #589, opened 2020-01-15
fix deprecated analyzer

kevmoo:


kevmoo:
Fixed, @natebosch

json_serializable/lib/src/json_key_utils.dart
```diff
@@ -68,7 +68,7 @@ JsonKey _from(FieldElement element, JsonSerializable classAnnotation) {
       // TODO(kevmoo): Support calling function for the default value?
       badType = 'Function';
     } else if (!reader.isLiteral) {
-      badType = dartObject.type.name;
+      badType = dartObject.type.getDisplayString();
```
jakemac53: Lets be consistent about `.element.name`.

kevmoo: fixed

jakemac53: The others too :)

kevmoo:
> hmm, previously this was a cascade which should have had no effect? This might have actually changed the behavior.

The cascades were copy-paste mistakes. Look at the diff now. Much better!

PR #602, opened 2020-02-12
Export DurationHelper

pdf:
This appears to be the only core helper that is not exported for public usage.

### DIFF
--- a/json_serializable/lib/type_helper.dart
+++ b/json_serializable/lib/type_helper.dart
@@ -7,6 +7,7 @@ export 'src/type_helper.dart' show TypeHelperContext, TypeHelper;
 export 'src/type_helpers/big_int_helper.dart';
 export 'src/type_helpers/convert_helper.dart';
 export 'src/type_helpers/date_time_helper.dart';
+export 'src/type_helpers/duration_helper.dart';
 export 'src/type_helpers/enum_helper.dart';
 export 'src/type_helpers/iterable_helper.dart';
 export 'src/type_helpers/json_converter_helper.dart';


pdf:
In a related question, should `TypeHelperContextWithConfig` and `TypeHelperContextWithConvert` also be exported? I don't know if the intention is for the helpers to be callable independently, but various helpers require these interfaces.

kevmoo:
> In a related question, should `TypeHelperContextWithConfig` and `TypeHelperContextWithConvert` also be exported? I don't know if the intention is for the helpers to be callable independently, but various helpers require these interfaces.

You ask a good question!

The idea was that `TypeHelperContext` would be simple – and be usable by most `TypeHelper` implementations.

Some require more information – hence these other classes.

The trick: if you're creating your own generation system (that's not `json_serializable`) you'd have to provide these richer types – which may be tough unless you implement all of the logic in this package.

If you're just coding to `json_serializable`, then you only need those other types if they have info you need.

pdf:
> you'd have to provide these richer types

When you say this, are you referring to the helpers that require this extra information, or to the contexts that provide it? If the latter, that can't be done currently because those additional context interfaces are not exported, so they can't be referenced in an `implements` statement, hence my question.

kevmoo:
Both! You'll need to create instances of them and provide them.

If you want to create a PR to export them, I'm happy to merge it!

pdf:
When updating the changelog, should I use a version of `unreleased`, or the next patch version? And would you like me to add the changelog entry for this to the new PR?

kevmoo:
Unreleased is fine. I'll take care of picking a version. :-)

kevmoo:
v3.3.0 is out now w/ this change!

PR #603, opened 2020-02-13
Export TypeHelperContextWithConfig in json_serializable/type_helper.dart

pdf:
Looks like this is the only one not exported - `TypeHelperContextWithConvert` is exported implicitly from `convert_helper.dart`.

refs #602


kevmoo:
v3.3.0 is out now w/ this change!

PR #604, opened 2020-02-13
Ignore implicit_dynamic_parameter in generated code

### LINKED ISSUE #557: Generate code that supports `implicit-dynamic: false`
Here's the generated code

```
Result _$ResultFromJson(Map<String, dynamic> json) {
  return Result(
    total: json['total'] as String,
    page: json['page'] as int,
    pages: json['pages'] as int,
    tv_shows: (json['tv_shows'] as List)
        ?.map((e) => e == null ? null : Tv.fromJson(e as Map<String, dynamic>))
        ?.toList(),
  );
}

```
And here's the error.

```
 error • Missing parameter type for 'e' at lib/demo.g.dart:15:16 • implicit_dynamic_parameter
```

Link: https://circleci.com/gh/trevorwang/retrofit.dart/150?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

MisterJimson:
**Done and looking for feedback on**
- Resolved #557 by ignoring specific warning in `json_serializable` generated code
- Updated `example` generated code
- Added `implicit-dynamic: false` to `example` to demo, let me know if this should be removed

**Need advisement**
- Updating tests. I have fixed 1 test out of ~40 that fail because the expected output is no longer correct. I'm happy to update them (see the one I updated), but I am wondering if we would prefer to do a contains check instead of an exact match check (which would make all these tests pass without having to update the expected outputs). Or if the inner test code should strip out the warning before comparing. May want to consider future lints we need to disable (NNDB as one example, other Strong Mode stuff, etc)

### DIFF
--- a/json_serializable/lib/src/utils.dart
+++ b/json_serializable/lib/src/utils.dart
@@ -256,3 +256,7 @@ extension DartTypeExtension on DartType {
       element.library == null ||
       element.library.typeSystem.isAssignableTo(this, other);
 }
+
+/// Comment to disable analysis errors not currently supported in generated
+/// output. Added to the top of all generated files.
+const ignoredWarningsComment = '// ignore_for_file: implicit_dynamic_parameter';
--- a/json_serializable/lib/src/utils.dart
+++ b/json_serializable/lib/src/utils.dart
@@ -256,7 +256,3 @@ extension DartTypeExtension on DartType {
       element.library == null ||
       element.library.typeSystem.isAssignableTo(this, other);
 }
-
-/// Comment to disable analysis errors not currently supported in generated
-/// output. Added to the top of all generated files.
-const ignoredWarningsComment = '// ignore_for_file: implicit_dynamic_parameter';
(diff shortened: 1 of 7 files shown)


json_serializable/lib/src/utils.dart
```diff
@@ -256,3 +256,7 @@ extension DartTypeExtension on DartType {
       element.library == null ||
       element.library.typeSystem.isAssignableTo(this, other);
 }
+
+/// Comment to disable analysis errors not currently supported in generated
+/// output. Added to the top of all generated files.
+const ignoredWarningsComment = '// ignore_for_file: implicit_dynamic_parameter';
```
kevmoo: Put this in `json_serializable_generator` and make it private!

MisterJimson:
I don’t see it as a blocker. The ignore comment only gets added once in the file no matter how many classes you annotate in the file. 

That issue would improve the implementation, but tons of libraries add this to generated output as is.

json_serializable/lib/src/utils.dart
MisterJimson: Done

kevmoo:
See https://github.com/dart-lang/json_serializable/pull/608

PR #614, opened 2020-02-22
Fix generated documentation

kevmoo:


json_serializable/README.md
```diff
@@ -91,26 +91,26 @@ is generated:
...
-[JsonKey.nullable]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/nullable.html
-[JsonKey.defaultValue]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/defaultValue.html
-[JsonKey.disallowNullValue]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/disallowNullValue.html
-[JsonKey.fromJson]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/fromJson.html
-[JsonKey.ignore]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/ignore.html
-[JsonKey.name]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/name.html
-[JsonKey.required]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/required.html
-[JsonKey.toJson]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/toJson.html
-[JsonKey.unknownEnumValue]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/unknownEnumValue.html
+[JsonSerializable.anyMap]: https://pub.dev/documentation/json_annotation/3.0.1/json_annotation/JsonSerializable/anyMap.html
```
jakemac53: why do we want a hard coded number here?

kevmoo: Because it aligns with the version of `json_annotation` that's compatible – at least that was my thought.

We could imagine someone looking at a previous version of `json_serializable` – and wanting to be pointed to a compatible version.

kevmoo:
CI broken by https://github.com/dart-lang/sdk/issues/40744

json_serializable/README.md
jakemac53: Ah ok that makes sense - but I think we would also need a test that ensures this is kept up to date with the current version number.

kevmoo: We do have a test, but it depends on the cache being invalidated on travis – I think

PR #615, opened 2020-02-22
Enable and fix 3 lints

kevmoo:


kevmoo:
> A lot of the bad strings are modified or deleted... are we concerned about that?

I think we're over testing these – I just did a copy-paste from the "bad string" index. We've had no issues here. I don't expect any.

...the HTML cases, for instance, aren't even a concern.

kevmoo:
Travis is failing due to https://github.com/dart-lang/sdk/issues/40744

kevmoo:
I wanted to enable the lint without having to wire through all of the crazy
ignores – at least that was my thought

kevmoo:
@jakemac53 – honestly, I'll just get rid of the 80-char line lint. Too much noise

PR #624, opened 2020-03-25
Add an explicit type on enumMaps

natebosch:
This works around a bug in the NNBD SDKs. It should not be necessary in
the long term, but also shouldn't hurt anything.

Work around https://github.com/dart-lang/sdk/issues/41180

### DIFF
--- a/json_serializable/lib/src/type_helpers/enum_helper.dart
+++ b/json_serializable/lib/src/type_helpers/enum_helper.dart
@@ -81,7 +81,8 @@ String _enumValueMapFromType(DartType targetType) {
           '${jsonLiteralAsDart(e.value)},')
       .join();
 
-  return 'const ${_constMapName(targetType)} = {\n$items\n};';
+  return 'const ${_constMapName(targetType)} = '
+      '<${targetType.element.name}, dynamic>{\n$items\n};';
 }
 
 const _enumDecodeHelper = r'''
--- a/json_serializable/test/default_value/default_value.g.dart
+++ b/json_serializable/test/default_value/default_value.g.dart
@@ -88,7 +88,7 @@ T _$enumDecodeNullable<T>(
   return _$enumDecode<T>(enumValues, source, unknownValue: unknownValue);
 }
 
-const _$GreekEnumMap = {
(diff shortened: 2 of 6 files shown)


kevmoo:
@natebosch – I can wait publishing `3.3.0` until this lands...if you like

...but I'm guessing you'll need to regenerate code

natebosch:
> I'm guessing you'll need to regenerate code

Yeah probably. Wanted to get this to you ASAP so you could decide if you want to run with it or not.

Another option is if we know the types of the values we could use a real type instead of `dynamic` and add it as an additional generic on the other methods. I don't know if we always have a type for that.

Yet another option would be to ignore it and hope that the SDK gets fixed soon. I mostly had this change because when this fails it's really annoying to figure out what's happening as @jakemac53 and I discovered.

kevmoo:
> > I'm guessing you'll need to regenerate code
> 
> Yeah probably. Wanted to get this to you ASAP so you could decide if you want to run with it or not.
> 
> Another option is if we know the types of the values we could use a real type instead of `dynamic` and add it as an additional generic on the other methods. I don't know if we always have a type for that.
> 
> Yet another option would be to ignore it and hope that the SDK gets fixed soon. I mostly had this change because when this fails it's really annoying to figure out what's happening as @jakemac53 and I discovered.

Where are we hitting this w/ the migration?

jakemac53:
I ran into this because I happened to be using the NNBD sdk when trying to build `build_modules`. 

All json serializable things that use enums are broken in the NNBD sdk.

kevmoo:
> I ran into this because I happened to be using the NNBD sdk when trying to build `build_modules`.
> 
> All json serializable things that use enums are broken in the NNBD sdk.

This is going to blow up all tests that validate build when it's released.

I'd rather wait for the SDK to get fixed. Doesn't seem urgent at all now.

natebosch:
> I'd rather wait for the SDK to get fixed. Doesn't seem urgent at all now.

Fair enough. We'll just need to make sure we're _not_ using the NNBD SDK locally when we try to do builds until this is fixed.

PR #627, opened 2020-04-02
Ignore use of CastError - deprecated in Dart 2.8.0-dev.18

kevmoo:


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

natebosch:
I didn't realize some of these were test/example only. Anything outside of lib code could be moved to `TypeError` if you want.

kevmoo:
> I didn't realize some of these were test/example only. Anything outside of lib code could be moved to `TypeError` if you want.

Not if we want to keep testing on old SDKs, though – right?

natebosch:
> Not if we want to keep testing on old SDKs, though – right?

Correct. I'm not sure how critical it is that the `example/` directory runs on old SDKs...

PR #632, opened 2020-04-22
Remove tests that fail with latest analyzer

kevmoo:
The latest analyzer doesn't allow ignoring errors


kevmoo:
I will create an issue (once this lands) to add back these tests in a way that doesn't use source code that'll cause errors

PR #650, opened 2020-06-04
Ignore deprecated member usage from pkg:analyzer

kevmoo:
DRY up use of type.isObject


json_serializable/lib/src/shared_checkers.dart
```diff
@@ -62,6 +62,9 @@ String asStatement(DartType type) {
   return ' as $typeCode';
 }
 
+// ignore: deprecated_member_use
```
natebosch, line 65: Is there an issue filed for the migration?

kevmoo: Wasn't planning on it. When we want to roll forward our analyzer dependency (and these members go away) we'll have to fix 'em

PR #668, opened 2020-07-11
Add new builder for each supported type

kevmoo:
Make it easy to validate basic behavior of each type in isolation


kevmoo:
This will greatly help implementing null safety, since it will be easy to create both null and non-nullable versions of each test without coding everything by hand.

PR #693, opened 2020-08-19
Add ignoreDecode/ignoreEncode support

### LINKED ISSUE #569: Implement JsonKey - readonly, writeonly
We have a class that contains fields that we need to fetch from the server but we can't send back to the server. As such we want to be able to define a field as being readonly and for symmetries sake we should have the ability to mark a field as write only.

Read only

```dart
class Team
{
@JsonKey(readOnly: true)
 List<User> members;
}
```

The `readonly` argument would cause the json serializer to suppress the generation of the `members` field in the toJson function.

Write only

```dart
class Team
{
@JsonKey(writeOnly: true)
 List<User> members;
}
```

The `writeOnly` argument would cause the json serializer to suppress the generation of the `members` field in the fromJson function.

charles-toller:
Fixes #569 

This pull request adds the ability to annotate a field with `ignoreDecode` or `ignoreEncode` keys, which work similar to the `ignore` key, but only for one side of the process, allowing for keys that are decoded but not serialized, or serialized but not decoded.


charles-toller:
Travis build appears to be failing because json_serializable needs the update from json_annotation, which pub can't find. Should I split this into two pull requests?

FernandoUFS:
This functionality will be really helpful to me

edlman:
I would suggest to reuse existing field's attributes `toJson` and `fromJson` - ignore it when set to `null`. I use following workaround with `ignore` global function which always returns null and null values are not processed (by `includeIfNull: false`).
```
@JsonSerializable(includeIfNull: false)
class Model {
  @JsonKey(fromJson: ignore)
  int writeOnlyNumber;
  @JsonKey(toJson: ignore)
  String readOnlyText;
}

T? ignore<T>(dynamic _) => null;
```

a-wallen:
This functionality would be very helpful to me

a-wallen:
> I would suggest to reuse existing field's attributes `toJson` and `fromJson` - ignore it when set to `null`. I use following workaround with `ignore` global function which always returns null and null values are not processed (by `includeIfNull: false`).
> 
> ```
> @JsonSerializable(includeIfNull: false)
> class Model {
>   @JsonKey(fromJson: ignore)
>   int writeOnlyNumber;
>   @JsonKey(toJson: ignore)
>   String readOnlyText;
> }
> 
> T? ignore<T>(dynamic _) => null;
> ```

This is a perfect workaround

kevmoo:
More likely to go with something like https://github.com/google/json_serializable.dart/pull/1178/

PR #691, opened 2020-08-17
Allow annotated field with JsonKey.unknownEnumValue to be of type List

jpelgrim:
This fixes https://github.com/google/json_serializable.dart/issues/585 and allows us to annotate lists with a default enum value when we encounter an unknown value.

Example:
```dart
  @JsonKey(unknownEnumValue: YourEnum.yourUnkownValue)
  final List<YourEnum> lastOrder;
```
This is currently not possible.

PR #699, opened 2020-08-26
Allow JsonKey.unknownEnumValue to be used on fields of Set or Iterable

kevmoo:
Throw an error in more cases if unknownEnumValue is used on an
incompatible type

Add a number of new tests

Follow-up to https://github.com/google/json_serializable.dart/pull/691


kevmoo:
FYI @jpelgrim – turns out it was pretty easy to generalize your work to support `Iterable` and `Set`, too!

PR #739, opened 2020-11-04
Drop build_web_compilers

kevmoo:
Not needed
Speeds up build and test...a LOT


jakemac53 (approved):
Ya it looks like this wasn't caching builds which would be the only potential advantage but I would just use pub run test like this as well, it works well enough.

PR #740, opened 2020-11-06
Support null safety

kevmoo:
Fixes https://github.com/google/json_serializable.dart/issues/712


_test_yaml/test/src/build_config.g.dart
```diff
@@ -36,35 +34,29 @@ Map<String, dynamic> _$ConfigToJson(Config instance) => <String, dynamic>{
...
-        '${enumValues.values.join(', ')}');
+    throw ArgumentError(
+      'A value must be provided. Supported values: '
+      '${enumValues.values.join(', ')}',
+    );
   }
 
   final value = enumValues.entries
-      .singleWhere((e) => e.value == source, orElse: () => null)
+      .cast<MapEntry<T, Object>?>()
```
jakemac53, line 49: Might want to work on a better solution for this... for instance you could eagerly throw the argument error inside of `orElse` and not need the cast?

json_annotation/lib/src/checked_helpers.dart
```diff
@@ -125,7 +135,9 @@ class CheckedFromJsonException implements Exception {
         'CheckedFromJsonException',
         if (_className != null) 'Could not create `$_className`.',
         if (key != null) 'There is a problem with "$key".',
-        if (message != null) message,
-        if (message == null && innerError != null) innerError.toString(),
+        if (message != null)
+          message!
```
jakemac53, line 139: Use `{}` now that this got broken onto its own line

json_serializable/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
-## 3.5.1-dev
+## 4.0.0-nullsafety.1
 
+- Generates null-safe code.
 - Improved error messages for unsupported types.
 - `package:json_serializable/type_helper.dart`
   - Made the third parameter to `UnsupportedTypeError` positional (optional).
+- **BREAKING** `bool defaultProvided` arg added to `TypeHelper.deserialize`.
```
jakemac53, line 7: Also mention the removal of `nullable` options in favor of respecting the nullability of the types.

json_annotation/lib/src/json_serializable.dart
```diff
@@ -172,20 +172,11 @@ class JsonSerializable {
...
-
-  /// When `true` (the default), `null` fields are handled gracefully when
-  /// encoding to JSON and when decoding `null` and nonexistent values from
-  /// JSON.
-  ///
-  /// Setting to `false` eliminates `null` verification in the generated code,
-  /// which reduces the code size. Errors may be thrown at runtime if `null`
-  /// values are encountered, but the original class should also implement
-  /// `null` runtime validation if it's critical.
-  final bool nullable;
```
jakemac53, line 185: I am a bit unsure that this should really be dropped at this time. Legacy projects may want to update their deps like analyzer, and thus require the latest json_serializable, but then no longer have a way of annotating their types as nullable since they are legacy still?

Up to you ultimately.

json_serializable/CHANGELOG.md
```diff
@@ -51,6 +54,7 @@
 
 - Support properties where the getter is defined in a class with a corresponding
   setter in a super type.
+>>>>>>> master
```
jakemac53, line 57: cruft

kevmoo: oops!

json_annotation/lib/src/json_serializable.dart
kevmoo: Hrm. I left the entries in the constructor, but I removed the fields. So for the most common usage, folks aren't blocked.

We can always come back and publish a before-the-break version of json_serial if folks want it w/ latest analyzer

json_serializable/CHANGELOG.md
kevmoo: Right!

json_annotation/lib/src/json_serializable.dart
jakemac53: There is definitely a lot of appeal to not trying to support weird mixed version stuff.

I also noticed after I left this that you already throw if running on a non-opted in library, so if it is a clean break that SGTM.

json_annotation/lib/src/checked_helpers.dart
kevmoo: Cannot use `{}` within list comprehension syntax. :shrug:

json_annotation/lib/src/json_serializable.dart
kevmoo: ack

_test_yaml/test/src/build_config.g.dart
kevmoo: Hrm...good idea!

jakemac53 (approved):

PR #754, opened 2020-11-25
Add support for nested field names

rebaz94:
Hi
I added support for accessing nested field directly. and I included different examples and all the tests are passed, Please check it I don't know i miss something.

Thanks


example/.idea/example.iml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kevmoo, line 1: Please delete this file!

json_serializable/.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kevmoo, line 1: please delete this file

json_serializable/lib/src/nested_field_utils.dart
```diff
@@ -0,0 +1,23 @@
+class NestedUtils {
```
kevmoo, line 1: put these at the root of the file. No need to wrap them in a class as static functions

kevmoo (commented):
An interesting idea.

As with any big PR like this, It's good to start a discussion w/ maintainers before you submit – so no one wastes their time.

My big concern: this is a breaking change.

It's totally valid to have `@JsonKey(name: 'a.b')` today – you're adding meaning here which might break folks.

rebaz94:
Thanks for your reply. 
I am sorry not familiar with PR for something like this, I needed in my project and I see many issues so decided to make this and also community use it.

you right but I think is minor breaking not so many people use ```.``` in field names

kevmoo:
> Thanks for your reply.
> I am sorry not familiar with PR for something like this, I needed in my project and I see many issues so decided to make this and also community use it.
> 
> you right but I think is minor breaking not so many people use `.` in field names
> 
> sorry my english is not good

You're english is likely better than my ability to speak your native language. 😄 

We have to care about "minor breaks" because we don't want to harm others.

We could change the type of `name` to be Object – and allow `String` or `List<String>` – that'd be one option.

kevmoo:
...run `dartfmt -w .` on the whole repo. That should get rid of a lot of the failures

example/.gitignore
```diff
@@ -3,3 +3,4 @@
 .pub/
 build/
 pubspec.lock
+.idea/
```
kevmoo, line 6: No! These should be in your `~/.gitignore_global` file! 😄 

See https://dart.dev/guides/libraries/private-files

rebaz94: It already ignored but I don't know why :D. i will fix now

rebaz94:
> You're english is likely better than my ability to speak your native language. 😄
> 
> We have to care about "minor breaks" because we don't want to harm others.
> 
> We could change the type of `name` to be Object – and allow `String` or `List<String>` – that'd be one option.

Hhhh Thanks.

Using list for path I think its boring, We could add another option like ```JsonKey(path: 'field.name')``` and this way if user specifically provide **path**, that will used otherwise to default to use ```name``` option

example/.gitignore
rebaz94: I run ```dartfmt -w .``` on the root project but still a lot of failure, why?

kevmoo:
You missed other packages. You have to do it at the very root of the
repository

rebaz94:
Hi @kevmoo 
What about creating new field in JsonKey class in order to provide **path** for the json field?
```dart
@JsonKey(path: 'nested.path.here')
```

This way used there is not breaking anyone's code and default is using ```name``` property.
please let know if thats good, I can make some changes :)
Thanks

kevmoo:
Maybe? I think `List` is a better option, though. Then we don't get into
the parsing game.

rebaz94:
Here simple example, using both option

```dart
@JsonSerializable()
class Example {
  @JsonKey(path: ['info', 'firstName'])
  String firstName;

  @JsonKey(path: ['info', 'lastName'])
  String lastName;
}
```

using just string
```dart
@JsonSerializable()
class Example {
  @JsonKey(path: 'info.firstName')
  String firstName;

  @JsonKey(path: 'info.lastName')
  String lastName;
}
```

Maybe the second option is more readable and if there is more options provided to ```JsonKey``` the formatting code is much nicer than using list. also I think you don't need to handle ```name``` and ```path``` field separately, but for performance reason yes the list is ready we don't need any extra split

rebaz94:
@kevmoo in the pubspec of json_serializable the **json_annotation** added as a package dep, how to override with the updated version for testing purpose? the test failed because the new ```path``` field is not found

kevmoo:
You'll need to do a dependency_override w/ a path

...although, forgive me, I think I'll ask you to WAIT on this PR. I want to land null safety support first before I do any more changes to this package!

jeffaknine:
Could we get an update on this ?

alexandradeas:
First off, I ended up here looking for this feature, so good work on implementing it and I look forward to bringing this into my projects when this lands.

I did want to make a point on the API for this change though.

This is valid json
```lang=json
{
  "a.b.c": 1,
  "a": {
    "b": {
      "c": 2
    }
  }
}
```

At the moment `@JsonKey(name: "a.b.c")` wouldn't be accepted anyway, but I wonder if for the sake of supporting the full JSON spec it'd make sense to move this to a different parameter such as `path` to avoid ambiguity and potentially collisions

rebaz94:
Hi @kevmoo 
I implemented support for nested json key after you upgraded to null safety, please could you see this https://github.com/rebaz94/json_serializable.dart

also I added ```path``` field to ```@JsonKey```.  so that does **not** breaking existing code. 

Here example
```dart
@JsonSerializable()
class NestedFields {
  @JsonKey(path: 'name.firstName')
  final String firstName;

  @JsonKey(path: 'name.lastName')
  final String lastName;

  @JsonKey(path: 'age')
  final int age;

  @JsonKey(path: 'name.family.firstName')
  final String familyFirstName;

  @JsonKey(path: 'name.family.lastName')
  final String familyLastName;

  @JsonKey(path: 'location.name')
  final String locationName;

  @JsonKey(path: 'location.rating')
  final double rating;

  @JsonKey(path: 'location.travelTime')
  final Duration travelTime;

  NestedFields(
    this.firstName,
    this.lastName,
    this.age,
    this.familyFirstName,
    this.familyLastName,
    this.rating, {
    required this.locationName,
    this.travelTime = const Duration(days: 1),
  });
}
```

kevmoo:
I think a better path here would be to make `name` `Object?` and allow either a String or a List<String>.

Then it's not a big breaking change...

rebaz94:
I don't think its a breaking change. both option are available and working correctly with existing codebase and I am using this for more than 4 months.
Please could you tell me the breaking changes?

kevmoo:
@rebaz94 – sorry. It's not *strictly* a breaking change to add a field, but it does add complexity. We now have two ways to do the same thing.

Which means we need to test that `name` and `path` aren't both set, etc.

Also, your model doesn't handle `{'some.key':{'another.key': 'value'}}` which is valid JSON, etc.

rebaz94:
You right it add complexity but its not so much and its much cleaner that using ```name``` as ```Object```.

The model already working or I am missing something!
```dart
@JsonSerializable()
class Test {
  Test(this.someObject);

  @JsonKey(name: 'some.key')
  final SomeValue someObject;

  factory Test.fromJson(Map<String, dynamic> json) => _$TestFromJson(json);
  Map<String, dynamic> toJson() => _$TestToJson(this);
}

@JsonSerializable()
class SomeValue {
  SomeValue(this.value);

  @JsonKey(name: 'another.key')
  String value;

  factory SomeValue.fromJson(Map<String, dynamic> json) => _$SomeValueFromJson(json);
  Map<String, dynamic> toJson() => _$SomeValueToJson(this);
}
```

using ```path``` field its just make things easier so that does not create a new class for every sub object. when things complicated it much better to wrap to their own class.

anyway thanks!

IanLuo:
is this pr going to merge?

alexandradeas:
I'm not convinced by the current notation, it doesn't support valid cases of JSON well.

```
{
  "a": {
    "b.1": {
      "c": "foo"
    }
  }
}
```

Would not be supported from what I can see. Using a `List<String>` would solve this issue as `c` can be referenced via `@JsonKey(path: ["a", "b.1", "c"])`

rebaz94:
@alexandradeas really never used dot notation inside json fields but from what I see its pretty easy to chanage ```path``` to accept a list and at the same time support regular ```name``` parameter.

PR #768, opened 2020-12-08
Allow latest pkg:yaml

kevmoo:
Prepare to publish v1.0.4

### DIFF
--- a/checked_yaml/lib/checked_yaml.dart
+++ b/checked_yaml/lib/checked_yaml.dart
@@ -28,7 +28,9 @@ T checkedYamlDecode<T>(
   YamlNode yaml;
 
   try {
-    yaml = loadYamlNode(yamlContent, sourceUrl: sourceUrl);
+    yaml = loadYamlNode(yamlContent,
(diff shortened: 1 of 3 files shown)


checked_yaml/lib/checked_yaml.dart
```diff
@@ -27,8 +27,17 @@ T checkedYamlDecode<T>(
...
+    // noop
+  } else if (sourceUrl is Uri) {
+    sourceUri = sourceUrl;
+  } else {
+    sourceUri = Uri.parse(sourceUrl as String);
+  }
+
   try {
-    yaml = loadYamlNode(yamlContent, sourceUrl: sourceUrl);
+    yaml = loadYamlNode(yamlContent, sourceUrl: sourceUri);
```
natebosch, line 40: I would recommend pushing `Uri` up to callers.

kevmoo: will do on the next breaking changed – just not now

natebosch (approved):

PR #799, opened 2021-02-11
Prepare to release v4 of json_serializable

kevmoo:


json_serializable/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## 4.0.0-dev
+## 4.0.0
 
+- Requires Dart 2.11 or greater.
```
natebosch, line 3: Do you want to mention why you are using this hack? I'd be confused as an end user to see that you require an SDK that doesn't exist.

natebosch (approved):

json_serializable/CHANGELOG.md
natebosch: Also I think @jakemac53 may have wanted to use `2.11.99` over in `build`...

kevmoo:
ack

PR #811, opened 2021-02-25
Serialise dates to UTC

via-guy:
If a `DateTime` object is serialised for an HTTP request but it's not in UTC then the ISO8601 string will be invalid, as it won't contain the timezone (missing from the dart implementation of `DateTime`). This causes requests to either fail or use the server timezone, which may not be the same as that of the device.

I know that I better solution would be to fix the `DateTime` function `toIso8601String`, but this is an easy solution here.


via-guy:
@googlebot I signed it!

guykogus:
@googlebot I consent.

kevmoo:
I'm hesitant to take this PR, just because it changes the behavior of something that's been VERY stable for a long time.

I think I'd rather consider using `toIso8601String` in a breaking change...

Want to open an issue for that?

Issue #816, opened 2021-03-01
checked_yaml: generate better errors for null values


kevmoo:
`type 'Null' is not a subtype of type 'String' in type cast` could be better.

Not sure if we should parse the string or do something else.

PR #818, opened 2021-03-02
checked_yaml: improve handling of missing keys

kevmoo:
If a CheckedFromJsonException exception is caught for a key that is missing
in the source map, include that information in the message

Also correctly handle the case where CheckedFromJsonException.message
is `null`

Fixes https://github.com/google/json_serializable.dart/issues/816

### DIFF
--- a/checked_yaml/lib/checked_yaml.dart
+++ b/checked_yaml/lib/checked_yaml.dart
@@ -75,12 +75,18 @@ ParsedYamlException toParsedYamlException(
       innerError: exception,
     );
   } else {
-    final yamlValue = yamlMap.nodes[exception.key];
-
-    if (yamlValue == null) {
-      // TODO: test this case!
+    if (exception.key == null) {
       return ParsedYamlException(
-        exception.message!,
+        exception.message ?? 'There was an error parsing the map.',
+        yamlMap,
+        innerError: exception,
+      );
+    } else if (!yamlMap.containsKey(exception.key)) {
+      return ParsedYamlException(
+        [
+          'Missing key "${exception.key}".',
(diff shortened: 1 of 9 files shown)


kevmoo:
@jakemac53 – related to your feedback on https://github.com/dart-lang/pubspec_parse/pull/65

checked_yaml/lib/checked_yaml.dart
```diff
@@ -75,12 +75,18 @@ ParsedYamlException toParsedYamlException(
...
+        yamlMap,
+        innerError: exception,
+      );
+    } else if (!yamlMap.containsKey(exception.key)) {
       return ParsedYamlException(
-        exception.message!,
+        [
+          'Missing key "${exception.key}".',
+          if (exception.message != null) exception.message!,
+        ].join(' '),
```
jakemac53, line 89:
```suggestion
        'Missing key "${exception.key}".${exception.message ?? ''}',
```

jakemac53 (approved):

checked_yaml/lib/checked_yaml.dart
kevmoo: ...but we want a space between them? Then you need to do double wrapping of string literals, etc. I think the code I wrote here is more readable. Less efficient, but it's not like this is going to be perf-critical.

natebosch: 😈 

```suggestion
'Missing key "${exception.key}"' + ' ${exception.message ?? ''}'.trimRight();
```

kevmoo: image

PR #819, opened 2021-03-02
json_annotation: Fix a potential error with `checked: true`...

kevmoo:
...when `ArgumentError.message` is `null`

### DIFF
--- a/json_annotation/lib/src/checked_helpers.dart
+++ b/json_annotation/lib/src/checked_helpers.dart
@@ -104,18 +104,25 @@ class CheckedFromJsonException implements Exception {
         innerStack = null;
 
   CheckedFromJsonException._(
-    this.innerError,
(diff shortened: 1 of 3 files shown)


json_annotation/lib/src/checked_helpers.dart
```diff
@@ -104,18 +104,25 @@ class CheckedFromJsonException implements Exception {
         innerStack = null;
 
   CheckedFromJsonException._(
-    this.innerError,
+    Object innerError,
```
jakemac53, line 107: See the comment on the linked issue, it looks like this can be `Object this.innerError`? Weird but seems to work.

kevmoo: Will try!

jakemac53 (approved):

Issue #803, opened 2021-02-17
Generic argument factories not working for nullable properties with null-safety


JonathanPeterCole:
When using generic argument factories for properties with a nullable generic type, the generated code expects the value to be non-null.

#### Source file
```dart
import 'package:json_annotation/json_annotation.dart';

part 'nullable_generic.g.dart';

@JsonSerializable(genericArgumentFactories: true)
class NullableGenericExample<T> {
  NullableGenericExample({this.maybeNull});

  factory NullableGenericExample.fromJson(
          Map<String, dynamic> json, T Function(Object? json) fromJsonT) =>
      _$NullableGenericExampleFromJson<T>(json, fromJsonT);

  final T? maybeNull;
}
```

#### Generated File
The generated file attempts to call `fromJsonT` and `toJsonT` regardless of whether or not the property is null. In the case of `toJsonT`, this will fail to compile when using null-safety as `T?` is being assigned to `T`.
```dart
// GENERATED CODE - DO NOT MODIFY BY HAND

part of 'nullable_generic.dart';

// **************************************************************************
// JsonSerializableGenerator
// **************************************************************************

NullableGenericExample<T> _$NullableGenericExampleFromJson<T>(
  Map<String, dynamic> json,
  T Function(Object? json) fromJsonT,
) {
  return NullableGenericExample<T>(
    maybeNull: fromJsonT(json['maybeNull']),
  );
}

Map<String, dynamic> _$NullableGenericExampleToJson<T>(
  NullableGenericExample<T> instance,
  Object Function(T value) toJsonT,
) =>
    <String, dynamic>{
      'maybeNull': toJsonT(instance.maybeNull),
    };
```

#### Expected Output
Something like this would fix the compile errors and prevent `fromJsonT` and `toJsonT` being called when the value is null:
```dart
NullableGenericExample<T> _$NullableGenericExampleFromJson<T>(
  Map<String, dynamic> json,
  T Function(Object? json) fromJsonT,
) {
  return NullableGenericExample<T>(
    maybeNull: json['maybeNull'] != null ? fromJsonT(json['maybeNull']) : null,
  );
}

Map<String, dynamic> _$NullableGenericExampleToJson<T>(
  NullableGenericExample<T> instance,
  Object Function(T value) toJsonT,
) =>
    <String, dynamic>{
      'maybeNull': instance.maybeNull != null ? toJsonT(instance.maybeNull!) : null,
    };
```

## Flutter version output:
```
[...]

PR #837, opened 2021-03-16
Add null check for toJsonT

sooxt98:
Fix an issue for `The argument type 'T?' can't be assigned to the parameter type 'T'.` https://github.com/google/json_serializable.dart/issues/803

image


sooxt98:
@googlebot I signed it!

Issue #821, opened 2021-03-05
Null safety version on pub dev


nank1ro:
Hello,

I can't find a null safety version of json_serializable on pub dev.
Can you publish one please?
I've seen this release `json_serializable-v4.0.0-nullsafety.0` here on github.
I've tried to use it with:
```
json_serializable:
    git:
      url: git://github.com/google/json_serializable.dart.git
      ref: 'json_serializable-v4.0.0-nullsafety.0'
      path: 'json_serializable/'
```
but when I run `dart pub outdated --mode=null-safety`
I get back:
`json_serializable  ✗4.0.0-nullsafety.0 (git)  ✗4.0.0-nullsafety.0 (git)  ✗4.0.0-nullsafety.0 (git)  ✗4.0.0-nullsafety.0 (git) `
so this release doesn't support null-safety yet

nank1ro:
Right know I was able to use the null safety version using:
```
json_serializable:
    git:
      url: git://github.com/google/json_serializable.dart.git
      ref: 'json_serializable-v4.0.0-nullsafety.0'
      path: 'json_serializable/'
```

nank1ro:
I've checked the pubspec of the release mentioned above, and I found:
```
environment:
  # Keeping this <2.12.0 because the code is not null safe – yet!
  sdk: '>=2.11.99 <3.0.0'
```
So, it's not yet null safe

pparikhatliaison:
@nank1ro you are correct. It is not null safe but the code generated is, i have pending PR waiting to merge which which makes it completely operational.

z85510:
How much it takes to release new version with null-safety ? 
I'm stuck on some task which related to library 🤦🏽‍♂️

kevmoo:
@z85510 – this package generates null-safe code – as of v4

The code itself is not null safe, but that shouldn't affect your package.

kevmoo:
Ignore the `--mode=null-safety` bits for this package. This code is run outside of your code.

kevmoo:
Waiting on down-stream dependencies to update first. This shouldn't block anyone from using this package w/ null-safety, though – as long as you're just using it as a normal builder.

namchuai:
I think we also need json_annotation to be null-safety as well.

fryette:
@namchuai it's already null safety

kevmoo:
v4.1.0 is out!

PR #841, opened 2021-03-18
Migrate implementation of json_serializable to null-safety

kevmoo:
Fixes https://github.com/google/json_serializable.dart/issues/821

### DIFF
--- a/json_serializable/CHANGELOG.md
+++ b/json_serializable/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.0-dev
+
+- Implementation is now null-safe.
+
 ## 4.0.3
 
 - Correctly handle nullable values with `genericArgumentFactories`.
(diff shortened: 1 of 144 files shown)


json_serializable/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.0.0-dev
```
jakemac53, line 1: You might be able to consider not doing a breaking release here if this package is not intended to be used as a library for any reason?

jakemac53 (approved):

json_serializable/CHANGELOG.md
kevmoo: It can be used as a lib, but it's not widely supported.

I'll bump this back in a follow-up. I have other fixes I want to get in.

Issue #869, opened 2021-04-08
Support for nullable `toJson`, `fromJson` on a non-nullable field.


xsahil03x:
Creating a custom `nullable` `toJson`, `fromJson` and providing a `defaultValue` for a `non-nullable` field should work out of the box.

```dart
@JsonKey(
    toJson: ForeignIdTimePair.toList, // nullable
    fromJson: ForeignIdTimePair.fromList, // nullable
    defaultValue: <ForeignIdTimePair>[],
  )
  final List<ForeignIdTimePair> deletedForeignIds;
```

Currently it throws:
```dart
Error with `@JsonKey` on `deletedForeignIds`. The `fromJson` function `fromList` return type `List<ForeignIdTimePair>` is not compatible with field type `List<ForeignIdTimePair>`.

```

kevmoo:
Please show the impls of ForeignIdTimePair to/fromList

xsahil03x:
Hey @kevmoo, sure 
```dart
  ///
  static List<ForeignIdTimePair>? fromList(List? pairs) {
    return pairs?.map((it) {
      final pair = it as List;
      checkArgument(pair.length == 2, 'Invalid foreignIdTime pair');
      final foreignId = pair[0] as String;
      final time = DateTime.parse(pair[1] as String);
      return ForeignIdTimePair(foreignId, time);
    }).toList(growable: false);
  }

  ///
  static List<List>? toList(List<ForeignIdTimePair>? pairs) {
    return pairs
        ?.map((it) => [it.foreignID, it.time.toIso8601String()])
        .toList(growable: false);
  }
  ```

**Comment by kevmoo** [comment]
You cannot use a nullable function with a non-nullable type.

I have a PR out to fix the error message.

**Comment by xsahil03x** [comment]
@kevmoo yes I agree, but the the `fromJson` should work if we are providing a default value in `JsonKey`
```dart
  @JsonKey(
    fromJson: ForeignIdTimePair.fromList, // nullable
    defaultValue: <ForeignIdTimePair>[],
  )
  final List<ForeignIdTimePair> deletedForeignIds;
```

it should normally generate 
```dart
deletedForeignIds: ForeignIdTimePair.fromList(data as List?) ?? <ForeignIdTimePair>[];
```

Feel free to fix me if I am wrong here.

kevmoo:
@xsahil03x – I'll follow up with that one!

kevmoo:
[...]

PR #873, opened 2021-04-13
Include nullability information for generic class

kevmoo:
Fixes https://github.com/google/json_serializable.dart/issues/869


natebosch (approved):
I think the commit message could use more detail about how the code changes in this diff accomplish the goal.

kevmoo:
@natebosch – good tip. I included more details when I hit the squash button!

PR #894, opened 2021-05-14
Correctly decode Map<String,double> when the input has int literals

kevmoo:
Prepare to release v4.1.2

Also fix directive ordering, which is now enforced
Also add stable branch (v4_x) to test on CI
Also update generated CI code with latest mono_repo


kevmoo:
TBR @natebosch

natebosch (commented):
LGTM

PR #916, opened 2021-06-10
Correctly handle converters and default values

kevmoo:
A consistent model for handling converter-like things and default values.

Also simplified and DRY'd up a number of code paths.

### DIFF
--- a/json_serializable/lib/src/decode_helper.dart
+++ b/json_serializable/lib/src/decode_helper.dart
@@ -199,7 +199,11 @@ abstract class DecodeHelper implements HelperCore {
         defaultProvided: defaultProvided,
       )!;
 
-      return DefaultContainer.encode(value, defaultValue: defaultValue);
+      return DefaultContainer.encode(
+        value,
+        nullable: targetType.isNullableType,
+        defaultValue: defaultValue,
+      );
     }
 
     String value;
--- a/json_serializable/lib/src/decode_helper.dart
+++ b/json_serializable/lib/src/decode_helper.dart
@@ -227,16 +227,6 @@ abstract class DecodeHelper implements HelperCore {
             'effect because both `disallowNullValue` and `required` are set to '
             '`true`.');
       }
-
-      final deserializeConvertData = contextHelper.deserializeConvertData;
-
-      if (deserializeConvertData != null &&
-          !deserializeConvertData.nullableToAllowDefault) {
-        log.warning('The field `${field.name}` has both `defaultValue` and '
-            '`fromJson` defined which likely won\'t work for your scenario.\n'
-            'Instead of using `defaultValue`, set `nullable: false` and handle '
-            '`null` in the `fromJson` function.');
-      }
     }
     return value;
   }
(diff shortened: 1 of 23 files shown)


json_serializable/lib/src/decode_helper.dart
```diff
@@ -227,16 +227,6 @@ abstract class DecodeHelper implements HelperCore {
             'effect because both `disallowNullValue` and `required` are set to '
             '`true`.');
       }
-
-      final deserializeConvertData = contextHelper.deserializeConvertData;
-
-      if (deserializeConvertData != null &&
-          !deserializeConvertData.nullableToAllowDefault) {
-        log.warning('The field `${field.name}` has both `defaultValue` and '
```
kevmoo, line 235: We now support this case! Nice!

json_serializable/lib/src/default_container.dart
```diff
@@ -27,6 +29,10 @@ class DefaultContainer {
       return input._result(defaultValue);
     }
 
+    if (input is LambdaResult && defaultValue != null) {
```
kevmoo, line 32: `LambdaResult` is used by the converters – so we can catch it here and handle the default case!

json_serializable/lib/src/lambda_result.dart
```diff
@@ -2,17 +2,23 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
+import 'constants.dart';
+
 class LambdaResult {
   final String expression;
   final String lambda;
+  final String? asContent;
+
+  String get _fullExpression => '$expression${asContent ?? ''}';
```
kevmoo, line 12: Need `expression` to remain "pure" so we can do the `null` check.

json_serializable/lib/src/type_helper_ctx.dart
```diff
@@ -125,8 +125,6 @@ ConvertData? _convertData(DartObject obj, FieldElement element, bool isFrom) {
         'positional parameter.');
   }
 
-  var nullableToAllowDefault = false;
```
kevmoo, line 128: no longer needed!

json_serializable/lib/src/json_key_utils.dart
```diff
@@ -202,7 +202,7 @@ KeyConfig _from(FieldElement element, ClassConfig classAnnotation) {
 KeyConfig _populateJsonKey(
   ClassConfig classAnnotation,
   FieldElement element, {
-  required Object? defaultValue,
+  required String? defaultValue,
```
kevmoo, line 205: Turns out we can be more specific here!

json_serializable/lib/src/type_helper.dart
```diff
@@ -91,5 +92,5 @@ Object commonNullPrefix(
   Object unsafeExpression,
 ) =>
     nullable
-        ? '$expression == null ? null : $unsafeExpression'
+        ? ifNullOrElse(expression, 'null', '$unsafeExpression')
```
kevmoo, line 95: Have this pattern three times, make it a shared util.

natebosch (approved):

PR #960, opened 2021-08-06
Make compatible with freezed package

mkranus:
Freezed package requires the JsonKey annotation to be inserted into the constructor, which is then copied to the generated classes. But currently this use generates a warning, and this PR adds an additional TargetKind so the warning is no longer generated.


mkranus:
@googlebot I signed it!

vi-k:
@JsonSerializable
factory ...

-> The annotation 'JsonSerializable' can only be used on classes.

No TargetKind for factory.

vi-k:
See https://github.com/rrousselGit/freezed/issues/474

Issue #980, opened 2021-09-18
generated code is wrong for class has fromJson ctor and type params in a List/Map/Set


windrunner414:
test.dart
```dart
import 'package:json_annotation/json_annotation.dart';

part 'test.g.dart';

typedef JsonMap = Map<String, dynamic>;

class A<T> {
  final T value;

  const A(this.value);

  factory A.fromJson(JsonMap json) => A(json['value']);

  JsonMap toJson() => {'value': value};
}

@JsonSerializable()
class B {
  final List<A<int>> list;

  const B(this.list);

  factory B.fromJson(JsonMap json) => _$BFromJson(json);

  JsonMap toJson() => _$BToJson(this);
}

```
and the generated code is
```dart
// GENERATED CODE - DO NOT MODIFY BY HAND

part of 'test.dart';

// **************************************************************************
// JsonSerializableGenerator
// **************************************************************************

B _$BFromJson(Map<String, dynamic> json) => B(
      (json['list'] as List<dynamic>)
          .map((e) => A.fromJson(e as Map<String, dynamic>))
          .toList(),
    );

Map<String, dynamic> _$BToJson(B instance) => <String, dynamic>{
      'list': instance.list,
    };

```
`.map((e) => A.fromJson(e as Map<String, dynamic>))` is wrong, the correct code may be `.map((e) => A<int>.fromJson(e as Map<String, dynamic>))`

kevmoo:
@windrunner414 – are you having static errors with this?
Runtime errors?

What are the errors?

The type arguments are inferred in my tests...with no problems. I don't see an issue here!

windrunner414:
@kevmoo is a static error.Have you tried to run the above code?
```
      (json['list'] as List<dynamic>)
          .map((e) => A.fromJson(e as Map<String, dynamic>))
          .toList()
```
The list inferred type is List<A<dynamic>>

kevmoo:
@windrunner414 – what's the version of the Dart SDK you're on?

windrunner414:
@kevmoo I'm outside now, so I'm not sure the specific version number.But I think it's 2.16 pre.I use flutter master channel a few days ago.

windrunner414:
@kevmoo just paste the code to dartpad, then you can see the error
```dart
void main() {
  
}

typedef JsonMap = Map<String, dynamic>;

class A<T> {
  final T value;

  const A(this.value);

  factory A.fromJson(JsonMap json) => A(json['value']);

[...]

PR #981, opened 2021-09-18
fix field is a List/Map/Set of element has fromJson ctor and type params

windrunner414:
fix this issue https://github.com/google/json_serializable.dart/issues/980


windrunner414:
FieldWithFromJsonCtorAndTypeParams test failed.
Now generated code is `MyList<GeneralTestClass2, int>.fromJson`, before is `MyList.fromJson`

kevmoo:
You need to run `dart run build_runner build` to update some files – then you're good!

windrunner414:
oh I use the wrong email in git......

Issue #778, opened 2020-12-21
Option to generate encoder/decoder for enum?


PiN73:
Currently one can't decode/encode enum values directly from/to string.

For example, I have
```dart
enum State {
  @JsonValue('in_review')
  inReview,

  ...
}
```

and want to do this
```dart
String s = 'in_review';
State state = stateFromString(s);
```

I need either implement `stateFromString` manually (which will duplicate strings in @JsonValue) or make a wrapper class.

It would be great to have an option to generate from/to string converters by some annotation, e.g.
```dart
@JsonSerializable()
enum State {
  ...
```

PiN73:
Maybe the generator can also reuse this generated function (like `stateFromString` and `stateFromString`) in `fromJson`/`toJson` for classes, e.g. if there is
```dart
class Article {
  State state;
  ...
}
```
inside another file

PiN73:
Also it would be great to have one place for setting `unknownEnumValue`, e.g.
```dart
@JsonEnumSerializable(unknownEnumValue: State.unknown)
enum State {
  ...
```
which will be used as default in `stateFromString`/`stateFromString` functions with possibility to override it

LichKing-2234:
You can follow this.
https://github.com/google/json_serializable.dart/issues/653#issuecomment-903583793

kevmoo:
I have an idea for this, but we'll need to wait for Dart 2.14

kevmoo:
WIP: https://github.com/google/json_serializable.dart/pull/989

HJ29:
Is it possible to convert enum to json value?
eg,
``` dart
enum StatusCode {
  @JsonValue(200)
  success,
  @JsonValue('500')
  weird,
}
```
Sometime, I want to display the `200` value, but I cant find a way to decode `StatusCode.success` to json value `200`.
json_serializable use `_$EnumNameEnumMap` and `_$enumDecode` to decode, but these aren't exposed to public,
so I cant use same way to decode.

PR #989, opened 2021-09-27
Add @JsonEnum annotation

kevmoo:
Allows generating the associated helpers without requiring usage as a field
in a @JsonSerializable class

Fixes https://github.com/google/json_serializable.dart/issues/778

### DIFF
--- a/json_serializable/lib/src/json_part_builder.dart
+++ b/json_serializable/lib/src/json_part_builder.dart
@@ -6,6 +6,7 @@ import 'package:build/build.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:source_gen/source_gen.dart';
 
+import 'json_enum_generator.dart';
 import 'json_literal_generator.dart';
 import 'json_serializable_generator.dart';
 import 'settings.dart';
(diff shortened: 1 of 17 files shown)


json_serializable/lib/src/json_part_builder.dart
```diff
@@ -23,10 +24,70 @@ Builder jsonPartBuilder({
...
+        const JsonEnumGenerator(),
+      ]),
       const JsonLiteralGenerator(),
     ],
     'json_serializable',
     formatOutput: formatOutput,
   );
 }
+
+/// Allows exposing separate [GeneratorForAnnotation] instances as one
```
natebosch, line 38: I don't understand why we need this.

What does this do differently than

```dart
SharedPartBuilder(
  [
    JsonSerializableGenerator.fromSettings(settonds),
     const JsonEnumGenerator(),
     const JsonLiteralGenerator(),
   ],
   'json_serializable',
   formatOutput: formatOutput,
 );
 ```

kevmoo: It's weird, I know. I need them to be in the same generator so duplicates get de-duped.

natebosch: Can you expand the doc comment on the class to make that clear?

kevmoo: https://github.com/google/json_serializable.dart/pull/990

PR #992, opened 2021-09-28
enum cleanup

kevmoo:
- test: move enum bits to their own library
- utils: refactor logic around getting enum fields
- json_key_util: cleanup field rename logic
- Allow specifying rename logic for enum entries using @JsonEnum


kevmoo:
FYI @natebosch

Issue #396, opened 2019-02-26
Allow non-String map keys, if there exists a compatible converter


kevmoo:
`Map<Version, int>` should be doable...

kevmoo:
Or `Map<int, int>` for that matter.

bgetsug:
I'd really like to be able to use value objects in place of `String`s to make my `Map`s a bit more clear. When this feature is complete will I be able to do something like the following?

```dart
@JsonSerializable()
class Something {
  Map<DotSeparatedKey, int> mapOfInts;
}

class DotSeparatedKey extends StringValue {
  DotSeparatedKey(List<String> values) : super(values.join("."));
}

class StringValue extends Value<String> {
  StringValue(String value) : super(value);
}

class StringValueConverter implements JsonConverter<StringValue, String> {
  @override
  StringValue fromJson(String json) => StringValue(json);

  @override
  String toJson(StringValue object) => object.value;
}

@immutable
abstract class Value<T> {
  final T value;

  Value(this.value);

  @override
  bool operator ==(Object other) =>
      identical(this, other) ||
      other is Value &&
          runtimeType == other.runtimeType &&
          value == other.value;

  @override
  int get hashCode => value.hashCode;
}

```

kevmoo:
@bgetsug – right now, no. I've wired through support for known types.

Getting this generic for an arbitrary key type will be...interesting. 😄 I think I can pull it off, though.

fzyzcjy:
Seems that now it is implemented! YEAH!

```
enum MyEnum { apple, banana, orange }

@JsonSerializable()
class HelloEntity {
  List<String>? list;
  Map<int, int>? intIntMap;
  Map<int, Object>? intObjMap;
  Map<MyEnum, HelloEntity>? enumRecursiveMap;
  Map<int, MyEnum>? intEnumMap;

  HelloEntity();

  factory HelloEntity.fromJson(Map<String, dynamic> json) => _$HelloEntityFromJson(json);

  Map<String, dynamic> toJson() => _$HelloEntityToJson(this);
}
```

```
HelloEntity _$HelloEntityFromJson(Map<String, dynamic> json) {
  return HelloEntity()
    ..list = (json['list'] as List<dynamic>?)?.map((e) => e as String).toList()
    ..intIntMap = (json['intIntMap'] as Map<String, dynamic>?)?.map(
      (k, e) => MapEntry(int.parse(k), e as int),
    )
[...]

PR #997, opened 2021-10-04
[WIP] Automatically convert custom Maps key types

TimWhiting:
Before
```dart
Map<int, String> // works
IMap<int, String> // does not work
```

Custom toJson / fromJson map types can now have keys serialized to strings. This expects the to/fromJson function to be of the form String Function(Object?) and Object? Function(String).

Resolves: #396

@kevmoo

### DIFF
--- a/json_serializable/lib/src/type_helpers/json_helper.dart
+++ b/json_serializable/lib/src/type_helpers/json_helper.dart
@@ -11,10 +11,13 @@ import 'package:source_gen/source_gen.dart';
 import 'package:source_helper/source_helper.dart';
 
 import '../default_container.dart';
+import '../shared_checkers.dart';
 import '../type_helper.dart';
+import '../unsupported_type_error.dart';
 import '../utils.dart';
 import 'config_types.dart';
 import 'generic_factory_helper.dart';
+import 'to_from_string.dart';
 
 const _helperLambdaParam = 'value';
 
@@ -49,11 +52,13 @@ class JsonHelper extends TypeHelper<TypeHelperContextWithConfig> {
 
       toJsonArgs.addAll(
         _helperParams(
+          context,
           context.serialize,
           _encodeHelper,
           interfaceType,
           toJson.parameters.where((element) => element.isRequiredPositional),
           toJson,
+          isSerializing: true,
         ),
       );
     }
@@ -109,11 +114,13 @@ class JsonHelper extends TypeHelper<TypeHelperContextWithConfig> {
       final args = [
         output,
         ..._helperParams(
+          context,
           context.deserialize,
           _decodeHelper,
           targetType,
           positionalParams.skip(1),
           fromJsonCtor,
+          isSerializing: false,
         ),
       ];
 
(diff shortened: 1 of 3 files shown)


json_serializable/README.md
```diff
@@ -26,6 +26,8 @@ Given a library `example.dart` with an `Person` class annotated with
 [`JsonSerializable`]:
 
 ```dart
+import 'dart:convert';
```
kevmoo, line 29: please don't add things to the default example!

kevmoo (commented):
One thought to add. This PR needs real tests. I'm not sure what's being fixed here!

json_serializable/README.md
TimWhiting: Reverted the changes to the example, and added an integration test which should help to show the use-case.

TimWhiting:
Addressed your comments.

The point of this PR is to allow custom implementations of from/toJson with generics to have key values automatically serialized to strings (such as ints, doubles, datetime, and enums). Keys were being serialized properly for built in map types, but not for custom maps such as the immutable map type I'm using from the [fast_immutable_collections](https://pub.dev/packages/fast_immutable_collections) package. The approach I took was to reuse the key serialization logic that is used for regular maps also for generic types that have `String?` types instead of the regular `Object?` type in the toJsonK/fromJsonK functions.

i.e.
```dart
class CustomMap<K, V> {
  final Map<K, V> map;

  CustomMap(this.map);

  factory CustomMap.fromJson(
    Map<String, dynamic> json,
    K Function(String?) fromJsonK,
    V Function(Object?) fromJsonV,
  ) =>
      CustomMap(json.map<K, V>(
          (key, value) => MapEntry(fromJsonK(key), fromJsonV(value))));

  Map<String?, dynamic> toJson(
          String? Function(K) toJsonK, Object? Function(V) toJsonV) =>
      map.map((key, value) => MapEntry(toJsonK(key), toJsonV(value)));
}
```

FXschwartz:
@TimWhiting Would this make it possible to serialize Firebase Timestamps?

TimWhiting:
@FXschwartz 
This PR is for custom container types with non-string keys into a string (with the same support as built in containers)

My understanding is that Firebase Timestamps are encoded as a json object.
```javascript
{
  nanoseconds: 0,
  seconds: 1562524200
}
```

Are you asking for Firebase Timestamps as keys? If so, in order to be valid json you can serialize it to a string to make it a string key. Which is related but not what this PR is for. The current implementation only handles the same things that normal maps handle as keys (e.g. enums, int, DateTime, Uri).

If you had something else / more specific in mind maybe I can help you find the right issue for it.

FXschwartz:
@TimWhiting  Sorry for the delay in response! Yes, I'm talking about being able to define Firebase Timestamp as a key.

I was hoping your PR was at least a workaround to #1072 where if I'm using the [Firestore ODM](https://firebase.flutter.dev/docs/firestore-odm/defining-models) package, json_serializable can handle the parsing of a Firestore Timestamp.

Hopefully, that makes sense, and yes if you have any ideas that would be greatly helpful, but probably don't need to exist in this PR discussion.

kevmoo:
I'm going to close this out. Feel free to re-open if you'd like to keep cranking...

PR #1040, opened 2021-11-25
Updated the tostring for UnrecognizedKeysException & DisallowedNullVa…

### LINKED ISSUE #874: Override toString in MissingRequiredKeysException and DisallowedNullValueException
This is an enhancement request. toString() is currently not implemented in MissingRequiredKeysException and DisallowedNullValueException, so we have to explicitly handle the exception and grab the message field.

Ideally, the string will start with the class name. For example, "MissingRequiredKeysException: required keys are missing: name, id, blah"

@override
String toString() => 'MissingRequiredKeysException: $message';

vanlooverenkoen:
Fixed #874 

Improved the to string of some exceptions

### DIFF
--- a/json_annotation/lib/src/allowed_keys_helpers.dart
+++ b/json_annotation/lib/src/allowed_keys_helpers.dart
@@ -73,7 +73,7 @@ class UnrecognizedKeysException extends BadKeyException {
       : super._(map);
 
   @override
-  String toString() => message;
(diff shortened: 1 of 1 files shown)


NicolaVerbeeck (approved):

vanlooverenkoen:
@kevmoo any idea when we can get this on pub.dev?

Issue #888, opened 2021-05-11
Serialize multiple JSON key in a single variable


Softkeydel:
With GSON we can do
@SerializedName(value="data", alternate={"variations", "products"})
@Expose
 private Data data;

How to implement the same with
@JsonKey(name: "location")

kenan0620:
> 使用GSON，我们可以执行
> @SerializedName（value =“ data”，Alternative = {“ variations”，“ products”}）
> @expose
> 私有数据数据；
> 
> 如何使用
> @jsonkey（名称：“位置”）实现相同功能

me too.
i want more to one field mapping. 

but eg is can't

@JsonKey(name: 'entity')
@JsonKey(name: 'items')

  List<String> list;

k-paxian:
something like [that](https://github.com/k-paxian/dart-json-mapper#name-aliases-configuration)?

kenan0620:
> something like [that](https://github.com/k-paxian/dart-json-mapper#name-aliases-configuration)?

tks,you are right, but u dart_json_mapper lib is old,
reason:Because no versions of json_serializable match >3.5.1 <4.0.0 and json_serializable 3.5.1 depends on build >=0.12.6 <2.0.0, json_serializable ^3.5.1 requires build >=0.12.6 <2.0.0.
u can update lib at someday? tks

k-paxian:
> > something like [that](https://github.com/k-paxian/dart-json-mapper#name-aliases-configuration)?
> 
> tks,you are right, but u dart_json_mapper lib is old,
> reason:Because no versions of json_serializable match >3.5.1 <4.0.0 and json_serializable 3.5.1 depends on build >=0.12.6 <2.0.0, json_serializable ^3.5.1 requires build >=0.12.6 <2.0.0.
> u can update lib at someday? tks

actually it's opposite, `json_serializable 3.5.1 depends on build >=0.12.6 <2.0.0`  and dart_json_mapper depends on `build ^2.0.1`
so you can use [previous version](https://pub.dev/packages/dart_json_mapper/versions/2.0.9) to be compatible alongside latest `json_serializable`

kenan0620:
> > > something like [that](https://github.com/k-paxian/dart-json-mapper#name-aliases-configuration)?
> > 
> > 
> > tks,you are right, but u dart_json_mapper lib is old,
[...]

Issue #144, opened 2018-05-05
Consider support aliases for name: for @JsonKey


Reekoooo:
I can't  do this: 
@JsonKey(name: "text", alternate = {"full_text"})
  String text;

in **gson** I can do something like:
@SerializedName(value = "text", alternate = {"full_text"})
    public final String text;

kevmoo:
To be clear, you want the option of reading `test` or `full_text` when deserializing?

Reekoooo:
the attribute sometimes is _"text"_ and some times is _"full_text_" i want to pick automatically between the two by providing alternate or alias name for the attribute i want to deserialize .. like I showed in the actual issue I want the equivalent of 
 
@serializedname(value = "text", alternate = {"full_text"})
public final String text;

walkermanx:
**yes. i need this too much! **

walkermanx:
**### i will follow this issue. looking forward to  the coming updates solving it .**

kevmoo:
Happy to take a PR here.

Jerry805:
Is this supported now?

kevmoo:
> Is this supported now?

If any work happens here, this issue will be updated

PR #1045, opened 2021-11-29
Allow customizable "reading" of JSON maps

kevmoo:
Allows aliasing values to different names or computing a single
field value from any aspect of the source JSON map

Fixes https://github.com/google/json_serializable.dart/issues/144
Fixes https://github.com/google/json_serializable.dart/issues/888


json_annotation/lib/src/checked_helpers.dart
```diff
@@ -13,11 +13,22 @@ typedef _CastFunction<R> = R Function(Object?);
 T $checkedCreate<T>(
   String className,
   Map map,
-  T Function(S Function<S>(String, _CastFunction<S>) converter) constructor, {
+  T Function(
+    S Function<S>(
+      String,
+      _CastFunction<S>, {
+      Object? Function(Map, String)? readValue,
```
jakemac53, line 20: Should `readValue` also return a `T` here to give better static checking? In general the relationship here between the `_CastFunction` and `readValue` params is a bit murky to me.

kevmoo: Will investigate...

kevmoo: So this is a bit weird.

`readValue` is meant to be pre-converter. So it should return a JSON-ish result. that then goes through any associated converters, etc.

If one wants `readValue` to return the field type, they have to also create a converter.

I pondered doing the logic where `readValue` just takes over ALL conversion logic, too. But that gets tricky.

jakemac53: Ya, that is what I was wondering about. This would cover some number of use cases, where one or more json fields can be merged into a new json-like field, but its a bit indirect for others. For instance you might have people actually returning a Map from `readValue`, built from some other fields, and then deserializing that map into some class in the regular conversion function.

Ultimately I think this choice is probably fine, but I would clarify things in the documentation?

kevmoo: Added more docs. PTAL

natebosch: > I pondered doing the logic where `readValue` just takes over ALL conversion logic, too. But that gets tricky.

Tricky for users or for this library?

It feels like a simpler concept to document if `readValue` does all the work instead of coordinating with some other uncoupled concept with no static types in between...

kevmoo: It could be both. Imagine a case where I just want to rename a field `foo` -> `bar` but the type is something complex like `Map<int, List<Duration?>>?` – that's a bear to rewrite for the user.

jakemac53 (approved):

PR #1078, opened 2021-12-30
adding support for custom instance on fromjson

zerobytes:
This PR will allow the _$CLASSFromJson method to optionally receive an instance of the object at a second parameter.

Whenever You already have an instance of a given class you can use populate the properties of this instance instead of creating a new one and replacing multiple previous pointers to the old instance.

This way you can have a method like this, on instance level.
```dart

void fromJson(Map<String, dynamic> json) => _$CLASSFromJson(json, existingInstance: this);

```


kevmoo:
I get the desire here, but this adds a bit too much noise/complexity for a relatively niche use case.

I'd recommend forking the repo if you need this for your use case!

Issue #1081, opened 2022-01-04
Request for backpatch to json_serializable v3 to allow analyzer v1 (and source_gen v0.9 to allow dart_style v2)


evanweible-wf:
Hello! We're working on upgrading our dart ecosystem to the latest of all dependencies, and one area we're stuck on right now is json_serializable. Since v4 outputs null-safe code, upgrading to it requires migrating to null-safety at the same time. We have stayed on v3 so far, but doing so precludes us from upgrading to analyzer v1, which in turn restricts our ability to upgrade a lot of other dependencies.

We'd really like to be able to take a more iterative approach to these dependency upgrades and null-safe migrations. If json_serializable v3 allowed analyzer v1 and source_gen v0.9 allowed dart_style v2, we'd be able to upgrade to the next major of the following packages _without_ also having to migrate to null-safety:

- analyzer
- build and build_*
- checked_yaml
- dart_style
- logging
- mockito
- pubspec_parse

And probably some more that we haven't enumerated yet.

I've opened PRs on my forks of this repo and source_gen to validate that json_serializable v3 is compatible with analyzer v1 and source_gen v0.9 is compatible with dart_style v2:

- https://github.com/evanweible-wf/json_serializable.dart/pull/1 (Note: the `ensure_build` step times out on Dart Stable, and I have no idea why. It passes locally on Stable.)
- https://github.com/evanweible-wf/source_gen/pull/1

If those back patches seem reasonable to you, I can open those PRs against the main repos once a base branch for those back patches exist. Let me know if there are any concerns with making these changes!

@kevmoo

kevmoo:
I'd accept a PR against v3. I'll open a branch now you can send a PR against.

kevmoo:
Use https://github.com/google/json_serializable.dart/tree/v3_x

evanweible-wf:
Awesome, thanks Kevin! Do you you maintain source_gen, too? Does that back patch seem reasonable as well?

kevmoo:
@evanweible-wf – I also work on source_gen. Happy to help there.

evanweible-wf:
Thank you for the help Kevin!

PR #1082, opened 2022-01-04
[Backpatch to v3] Widen analyzer range to allow v1

evanweible-wf:
See #1081 for details.

### DIFF
--- a/json_serializable/pubspec.yaml
+++ b/json_serializable/pubspec.yaml
@@ -8,9 +8,9 @@ environment:
   sdk: '>=2.7.0 <3.0.0'
 
 dependencies:
-  analyzer: '>=0.39.0 <0.42.0'
-  build: '>=0.12.6 <2.0.0'
-  build_config: '>=0.2.6 <0.5.0'
+  analyzer: '>=0.39.0 <2.0.0'
+  build: '>=0.12.6 <3.0.0'
+  build_config: '>=0.2.6 <2.0.0'
 
   # Use a tight version constraint to ensure that a constraint on
   # `json_annotation` properly constrains all features it provides.
@@ -23,11 +23,11 @@ dependencies:
 
 dev_dependencies:
   build_runner: ^1.0.0
-  build_verify: ^1.1.0
+  build_verify: ^2.0.0
   collection: ^1.14.0
-  dart_style: ^1.2.0
-  logging: ^0.11.3+1
-  pub_semver: ^1.4.0
+  dart_style: ^1.0.0
+  logging: ^1.0.0
+  pub_semver: ^2.0.0
   source_gen_test: ^0.1.0
   test: ^1.6.0
-  yaml: ^2.1.13
+  yaml: ^3.0.0
--- a/json_serializable/pubspec.yaml
+++ b/json_serializable/pubspec.yaml
@@ -1,5 +1,5 @@
 name: json_serializable
-version: 3.5.1
+version: 3.5.2
 description: >-
   Automatically generate code for converting to and from JSON by annotating
   Dart classes.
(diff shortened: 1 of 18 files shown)


json_serializable/pubspec.yaml
```diff
@@ -8,9 +8,9 @@ environment:
   sdk: '>=2.7.0 <3.0.0'
```
kevmoo, line 8: Make sure to bump version here, too!

kevmoo (changes requested):

json_serializable/pubspec.yaml
evanweible-wf: Hmm, how should I handle this? I updated CI to run on 2.13.4 so that I could use the newer mono_repo and `dart` CLI, but I can't raise the minimum to that without opting into null safety. Maybe I should use 2.10 instead?

kevmoo: We did SDK constraint of `>=2.11.99 <3.0.0` as a flag of "latest that doesn't do null safety"

kevmoo:
@evanweible-wf – you'll need to drop `stable` from all CI, too!

kevmoo:
> @evanweible-wf – you'll need to drop `stable` from all CI, too!

Well, actually. Not sure. Hrm. It seems that the ensure_build bits are hanging on stable, though.

evanweible-wf:
Yeah, I couldn't figure that one out. It hangs during CI but runs fine locally for me on stable.

evanweible-wf:
I take that back - I cleared out the .dart_tool in _test_yaml and now I'm seeing the same thing and running just a regular build shows this error:
```
[SEVERE] json_serializable:json_serializable on test/src/build_config.dart:

Bad state: Unexpected diagnostics:
/Users/evanweible/.asdf/installs/dart/2.15.1/dart-sdk/lib/core/int.dart:120:18 - Operator declarations must be preceded by the keyword 'operator'.
/Users/evanweible/.asdf/installs/dart/2.15.1/dart-sdk/lib/core/int.dart:120:16 - Methods must have an explicit list of parameters.
/Users/evanweible/.asdf/installs/dart/2.15.1/dart-sdk/lib/core/int.dart:120:18 - A function body must be provided.
```

Which I recognize from here: https://github.com/dart-lang/sdk/issues/47191

We've worked around this internally with the pin on `meta` (which I had added to `_test_yaml`'s pubspec.yaml), but that alone doesn't seem to be enough. Looking into it now.

evanweible-wf:
Ok, I was misremembering. The pin on `meta` avoids a separate issue. The build error above is due to being on an outdated version of analyzer while on SDK 2.14+. The only way to get `_test_yaml` to resolve to analyzer v1 would be to update `pkg:checked_yaml` to allow `pkg:yaml` v3. The only potential problem there is with the dynamic `sourceUrl` parameter that was changed to be typed as `Uri`.

@kevmoo do you think I should pursue that option or would it be sufficient to drop the `_test_yaml` CI step on stable? Afaict, this isn't indicative of a real issue in any of the packages in this repo - just running into https://github.com/dart-lang/sdk/issues/47191 because of the combination of Dart SDK and `pkg:analyzer` versions.

kevmoo:
Dropping `_test_yaml` on stable sounds good to me (SGTM).

kevmoo:
Also borked on `checked_yaml` 🤷

evanweible-wf:
@kevmoo Ok I went ahead and changed the CI for `_test_yaml` and `checked_yaml` to only run on 2.13.4, but kept the others running on 2.13.4 and stable. I figured that should be fine since json_serializable is the only one being published, but let me know if I should try something else!

kevmoo (approved):

PR #1104, opened 2022-03-02
Use `flutter pub add` to add dependencies

domesticmouse:
PTAL @kevmoo

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -3,13 +3,20 @@ _This example assumes you're using a recent version of the Dart or Flutter SDK._
 To use [package:json_serializable][json_serializable] in your package, add these
 dependencies to your `pubspec.yaml`.
 
-```yaml
-dependencies:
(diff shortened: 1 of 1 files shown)


kevmoo:
@domesticmouse – there's a kinda weird test that validates the contents of the readme you'll want to check!

kevmoo (changes requested):
Let's drop this change. I want to be clear on the versions needed, etc. It matters.

Maybe add a note after the existing block saying one can use `dart pub add` etc to add the deps?

domesticmouse:
> I want to be clear on the versions needed, etc. It matters.

ELI5 time: 
  - Why should a normal developer care about exact versions, as opposed to using the latest usable version as `dart pub add` does automatically? 
  - What does the developer gain by taking these additional step? 
  - What do we gain from this required maintenance?

Issue #1115, opened 2022-03-14
Allow usage of prefix for the import of json_annotation.dart


sleepychaman:
Hi !
And before all thank you for this package !

I found an issue while using prefix for the  import of json_annotation.dart and working with enum.
Say I have a file : 

```dart
// user.dart
import 'package:json_annotation/json_annotation.dart' as prefix;

part 'user.g.dart'

@prefix.JsonSerializable()
class User {
   final Gender gender;
   final String name;
   User({required this.gender, required this.name});
}

enum Gender{male, female, other}
```
after running `flutter pub run build_runner watch --delete-conflicting-outputs`
the .g file look like this : 
```dart
// GENERATED CODE - DO NOT MODIFY BY HAND

// ignore_for_file: implicit_dynamic_parameter

part of 'user.g.dart';

// **************************************************************************
// JsonSerializableGenerator
// **************************************************************************

User _$UserFromJson(Map<String, dynamic> json) => User(
      name: json['name'] as String,
      gender: $enumDecode(_$GenderEnumMap, json['status']),
    );

Map<String, dynamic> _$UserToJson(User instance) => <String, dynamic>{
      'name': instance.name,
      'lastName': instance.lastName,
      'gender': _$GenderEnumMap[instance.gender],
    };

const _$GenderEnumMap = {
 Gender.male: 'male',
  Gender.female: 'female',
  Gender.other: 'other',
};
```
But this will throw `Error: Method not found: '$enumDecode'.`
  because `$enumDecode` is define in json_annotation.dart but with the prefix it should be called like that : `prefix.$enumDecode`.

I made a PR to add the prefix before some helpers of json_annotation :  #1116 

```
Flutter 2.10.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 097d3313d8 (3 weeks ago) • 2022-02-18 19:33:08 -0600
Engine • revision a83ed0e5e3
Tools • Dart 2.16.1 • DevTools 2.9.2
```

xmine64:
not working for me as well

PR #1116, opened 2022-03-14
Allow usage of prefix for the import of json_annotation.dart

sleepychaman:
issue #1115

### DIFF
--- a/json_serializable/lib/src/utils.dart
+++ b/json_serializable/lib/src/utils.dart
@@ -232,12 +232,14 @@ extension ExecutableElementExtension on ExecutableElement {
 }
 
 // Return the prefix used by json_annotation if there is one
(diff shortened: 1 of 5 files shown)


json_serializable/lib/src/utils.dart
```diff
@@ -229,3 +230,14 @@ extension ExecutableElementExtension on ExecutableElement {
     );
   }
 }
+
+// Return the prefix used by json_annotation if there is one
+String? getPrefixOfJsonAnnotationImport(TypeHelperContextWithConfig context) {
```
kevmoo, line 235: Why not return `''` – empty string when there is no prefix and add a `.` when there is.

Then you can drop all of the if checks all over.

Maybe?

sleepychaman: Yes, you re right it will be better !

kevmoo:
Going to close this out for the time being. Feel free to try again!

Issue #1032, opened 2021-11-15
[Feature request] FieldRename upper snake case


leoshusar:
It would be nice to have enum `FieldRename` for upper (screaming?) snake case.

vicdotdevelop:
+1 <3

PR #1118, opened 2022-03-23
Add support for screaming snake case #1032

zbarbuto:
Adds support for `SCREAMING_SNAKE_CASE` field renaming ( #1032 ).

Happy to call it whatever - have also seen it called "macro case", "constant case" or "upper snake case" but screaming snake is what is used in the [wikipedia article on snake case](https://en.wikipedia.org/wiki/Snake_case).

### DIFF
--- a/checked_yaml/pubspec.yaml
+++ b/checked_yaml/pubspec.yaml
@@ -24,3 +24,5 @@ dev_dependencies:
 dependency_overrides:
   json_annotation:
     path: ../json_annotation
+  json_serializable:
+    path: ../json_serializable
--- a/json_annotation/lib/src/json_serializable.dart
+++ b/json_annotation/lib/src/json_serializable.dart
@@ -22,7 +22,11 @@ enum FieldRename {
   snake,
 
   /// Encodes a field named `pascalCase` with a JSON key `PascalCase`.
-  pascal
+  pascal,
+
+  /// Encodes a field named `screamingSnakeCase` with a JSON key
+  /// `SCREAMING_SNAKE_CASE`
+  screamingSnake,
 }
 
 /// An annotation used to specify a class to generate code for.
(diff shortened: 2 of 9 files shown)


zbarbuto:
Hmm, not sure what changed? This was passing when I first made it.

zbarbuto:
@kevmoo I'm having some difficulty understanding why the build is failing as my local analysis is passing. Would you be able to give me a hand please?

> There's no constant named 'screamingSnake' in 'FieldRename'.

However, I've definitely added it [here](https://github.com/google/json_serializable.dart/pull/1118/files#diff-50e38e71613307c10f7c9162eafbd8042997915e00d3a098e12c3a869efb5db1R29)

Is it because I've modified both packages at the same time?

kevmoo:
@zbarbuto – you need to update pubspec.yaml to make sure it references the other package...

zbarbuto:
Do you mean by path or by manually bumping the versions?

I'm using path dependency overrides locally but I figured that shouldn't be committed back to the repo since it gets published?

kevmoo:
it's fine to checkin the path dependency while developing. Otherwise, CI won't pass!

zbarbuto:
Yeah, understood.

Sorry for the force-push commit noise have been trying to get passing using the same actions in our fork. I think we should be good now.

zbarbuto:
The [1 remaining failure](https://github.com/google/json_serializable.dart/runs/6070373577?check_suite_focus=true) seems unrelated to these changes:

>   the last disk(s) of this archive.
> unzip:  cannot find zipfile directory in one of /home/runner/dartsdk.zip or
>        /home/runner/dartsdk.zip.zip, and cannot find /home/runner/dartsdk.zip.ZIP, period.

zbarbuto:
Added override dependencies for `checked_yaml` to hopefully fix that last failure.

kevmoo (approved):
Well done!  Thanks!

PR #1135, opened 2022-04-21
Add JsonConverter(converters) option

### LINKED ISSUE #1072: Add json allowlist of specific types
Currently, json_serializable does not support custom JSON format which contains classes other than Map/String/num/...

An example would be Firestore, which on top of your typical String/bool/... may also have Timestamp/GeoPoint/DocumentReference as Map values.  

Meaning that if we're doing:

```dart
@JsonSerializable
class Foo {
  <...>
  final Timestamp timestamp;
}
```

then json_serializable will complain with the error:

```
Could not generate `fromJson` code for `timestamp`.
To support the type `Timestamp` you can:
* Use `JsonConverter`
  https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonConverter-class.html
* Use `JsonKey` fields `fromJson` and `toJson`
  https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/fromJson.html
  https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/toJson.html
package:cloud_firestore_odm_generator_integration_test/simple.dart:15:19
```

The problem is, this error is technically incorrect in this specific use-case since we don't need to encode/decode `Timestamp`. 

We could use JsonConverter/JsonKey, but that is not ideal since they can't be applied to the entire project.

## Proposal

We could add a way to globally tell json_serializable to not do anything when encountering certain types.

This could be a configuration inside the `build.yaml` file, such as:

```yaml
json_serializable:
  options:
     allowed-types:
         - name: Timestamp
            source: "package:cloud_firestore/cloud_firestore.dart"
```

Then, when json_serializable will encounter a property typed as `Timestamp`, the generated code will simply perform as cast instead of trying to call fromJson/toJson.

rrousselGit:
This implements what is described here https://github.com/google/json_serializable.dart/issues/1072#issuecomment-1104918300 (assuming this is desirable)

That should simplify a bit the work of having to reuse a few JsonConverter over and over by allowing folks to make a custom reusable annotation or write things like:

```dart
const myConverters = [DocumentReferenceConverter()];

@JsonSerializable(converters: myConverters);
```

fixes #1072


rrousselGit:
I had to separate ClassConfig and JsonSerializable for this (because ClassConfig receives DartObjects for the converters but JsonSerializable receives a JsonConverter).

Hopefully that's alright.

kevmoo:
> I had to separate ClassConfig and JsonSerializable for this (because ClassConfig receives DartObjects for the converters but JsonSerializable receives a JsonConverter).
> 
> Hopefully that's alright.

Ha! We'll start by seeing if the tests pass. 😁

rrousselGit:
The CI just wanted a dependency_overrides for the annotation 😄

json_serializable/test/kitchen_sink/kitchen_sink.g_exclude_null.dart
```diff
@@ -196,13 +196,12 @@ class KitchenSink implements k.KitchenSink {
   }
 }
 
-@JsonSerializable(
-  includeIfNull: false,
-)
-// referencing a top-level field should work
-@durationConverter
```
kevmoo, line 203: I'd like to keep one example of using a constant. Do we still have one?

json_serializable/lib/src/type_helpers/config_types.dart
```diff
@@ -38,41 +39,20 @@ class KeyConfig {
 /// configuration.
 ///
 /// Values are all known, so types are non-nullable.
-class ClassConfig implements JsonSerializable {
-  @override
+class ClassConfig {
```
kevmoo, line 42: This refactoring is my least favorite change.

Did you ponder just leaving `converters` empty and adding a property here?

json_annotation/lib/src/json_serializable.dart
```diff
@@ -6,6 +6,7 @@ import 'package:meta/meta_meta.dart';
 
 import 'allowed_keys_helpers.dart';
 import 'checked_helpers.dart';
+import 'json_converter.dart';
```
kevmoo, line 9: Need to bump the minor version of this package. Add a changelog entry.

json_serializable/pubspec.yaml
```diff
@@ -36,3 +36,7 @@ dev_dependencies:
   test_descriptor: ^2.0.0
   test_process: ^2.0.0
   yaml: ^3.0.0
+
+dependency_overrides:
```
kevmoo, line 40: Need to bump the version range on `json_annotation` to be "tight" w/ what you change in the json_annotation package

kevmoo: also changelog!

kevmoo (changes requested):

rrousselGit:
I'm not quite sure how to fix that last CI error

One test seems to complain if I don't update the readme with 4.6.0 links. But the markdown checked complains that 4.6.0 links are 404 (which is logical since it isn't published yet)

kevmoo:
ignore the markdown errors. That's expected when things are revved!

kevmoo (approved):
Sweet!

Issue #822, opened 2021-03-06
Allow nullable JsonConverter's to work with both nullable and non-nullable types


shilangyu:
- Dart version: `Dart SDK version: 2.12.0 (stable) (Thu Feb 25 19:50:53 2021 +0100) on "linux_x64"
`
- json_serializable version: `4.0.2`

When creating a custom `JsonConverter` that handles nullable types, it will not be applied to non-nullables. Example, consider this:

```dart
import 'package:json_annotation/json_annotation.dart';

part 'mydata.g.dart';

class ForceUtcDateTime implements JsonConverter<DateTime?, String?> {
  const ForceUtcDateTime();

  @override
  DateTime? fromJson(String? json) => json == null
      ? null
      : DateTime.parse('$json${json.endsWith('Z') ? '' : 'Z'}');

  @override
  String? toJson(DateTime? json) => json?.toIso8601String();
}

@JsonSerializable()
@ForceUtcDateTime()
class MyData {
  final DateTime created;
  final DateTime? updated;

  const MyData({required this.created, this.updated});

  factory MyData.fromJson(Map<String, dynamic> json) => _$MyDataFromJson(json);
  Map<String, dynamic> toJson() => _$MyDataToJson(this);
}
```

A custom `DateTime` converter is created that handles `DateTime?`. The generated code is as follows:

```dart
MyData _$MyDataFromJson(Map<String, dynamic> json) {
  return MyData(
    created: DateTime.parse(json['created'] as String),
    updated: const ForceUtcDateTime().fromJson(json['updated'] as String?),
  );
}

Map<String, dynamic> _$MyDataToJson(MyData instance) => <String, dynamic>{
      'created': instance.created.toIso8601String(),
      'updated': const ForceUtcDateTime().toJson(instance.updated),
    };
```

I would expect both `created` and `updated` to be handled by my `ForceUtcDateTime` converted, not just `updated`

If this is intended, I would like to know what is the expected way of handling such cases.

Ludonope:
Having similar issue with a type in a list

```dart
// my_data.dart
@JsonSerializable()
@MyTypeConverter()
class MyData {
  late MyType nonNullable;
  MyType? yesNullable;
  List<MyType>? someList;

  const MyData();

  factory MyData.fromJson(Map<String, dynamic> json) => _$MyDataFromJson(json);
  Map<String, dynamic> toJson() => _$MyDataToJson(this);
}
```

```dart
// my_data.g.dart
[...]

PR #1136, opened 2022-04-21
Support using non-nullable JsonConverter on nullable properties

rrousselGit:
This allows using `JsonConverter<int, ...>` on `int? value`.

This also gracefully handles `null` JSON when using `JsonConverter<..., int>` on `Object? value`, such that `null` is returned instead of throwing

fixes https://github.com/google/json_serializable.dart/issues/822


json_serializable/lib/src/type_helpers/json_converter_helper.dart
```diff
@@ -49,6 +62,19 @@ class JsonConverterHelper extends TypeHelper {
 
     final asContent = asStatement(converter.jsonType);
 
+    if (!converter.jsonType.isNullableType && targetType.isNullableType) {
+      // Hacky way to hoist the expression to avoid casting
+      // Ideally we should be able to cleanly declare a variable in the
+      // parent code-block, instead of using (){}()
+      return '''
+() {
```
kevmoo, line 70: Can you "DRY" up this expression with the one above? Put it in a private, static helper?

json_serializable/test/generic_files/generic_class.g.dart
```diff
@@ -39,27 +39,42 @@ GenericClassWithConverter<T, S>
...
           ..fieldInt = json['fieldInt'] as int?
-          ..fieldT = _SimpleConverter<T?>()
-              .fromJson(json['fieldT'] as Map<String, dynamic>)
-          ..fieldS = _SimpleConverter<S?>()
-              .fromJson(json['fieldS'] as Map<String, dynamic>)
-          ..duration = const _DurationMillisecondConverter.named()
-              .fromJson(json['duration'] as int?)
-          ..listDuration = const _DurationListMillisecondConverter()
-              .fromJson(json['listDuration'] as int?);
+          ..fieldT = () {
```
kevmoo, line 42: Ugh. This looks pretty ugly here.

Maybe we should generate a `_$SomeKindOfHelper` top-level function?

rrousselGit: Would this involve sharing the helper if used multiple time? If so, I don't see an obvious solution for it.

kevmoo: There's a feature for that...to only generate a thing once.

rrousselGit: Could you give me a hint on how to do that? I see `context.addMember`, but it doesn't seem to generate only once.

kevmoo: addMember SHOULD only add once if the content is identical.

rrousselGit: Done, that should be better now.

json_serializable/lib/src/type_helpers/json_converter_helper.dart
```diff
@@ -49,6 +67,24 @@ class JsonConverterHelper extends TypeHelper {
 
     final asContent = asStatement(converter.jsonType);
 
+    if (!converter.jsonType.isNullableType && targetType.isNullableType) {
+      const converterFromJsonName = r'_$JsonConverterFromJson';
```
kevmoo, line 71: Uh....what if there are multiple different converters used in the same library?

rrousselGit: That's fine because the helper is independent from the converter. The helper is a bit more complex than necessary (generic helper with a callback parameter) to make it reusable between converters.

That's the only way I found to prevent name clash and handle fancy cases like:

```dart
const converter = MyConverter();

...
@converter
final int field;
```

kevmoo: Riiiiight!

kevmoo (approved):

kevmoo (changes requested):
Gah! Just thought of this. Create just one function. Call it `$convertOrNull`. There is zero benefit to have two identical functions with swapped generic arguments – right?

rrousselGit:
Good call, I didn't realise

rrousselGit:
Actually, I'm not sure if that's a meaningful improvement

Only one of the two functions (fromJson) is doing a cast, so their body has a difference. We always do the cast, but that'd negatively impact toJson which doesn't need one.

kevmoo (approved):

kevmoo (changes requested):
Please update the version to 6.3.0-dev and add a changelog entry

We can see if we can land you other change here, too!

rrousselGit:
Done

kevmoo (approved):

hasimyerlikaya:
Hi,

I have problem with this merge.
I use two converter for a custom type wich name is Ref.
I can't generate g.dart file because there is an error.

"Found more than one matching converter for `Ref?`."

I don't want to add annotation top of nullable field every time.
How can I use both converter with class level annotation?

**Non-Nullable Converter**
```
class RefResultConverter implements JsonConverter<Ref, dynamic> {
  const RefResultConverter();

  @override
  Ref fromJson(dynamic json) {
    if (json != null) {
      if (json["@ref"] != null) {
        final refResult = RefResult.fromJson(json["@ref"] as Map<String, dynamic>);
        return refResult.asRef();
      } else if (json["ref"] != null && json["ref"]["collection"] != null) {
        final collection = json["ref"]["collection"] as String;
        final id = json["id"] as String;
        return Ref(Collection(collection), id);
      }
    }
    return Ref.empty();
  }

  @override
  Ref toJson(Ref ref) {
    if (ref.id == null || ref.id == "") {
      return Ref.empty();
    }

    return ref;
  }
}
```

**Nullable Converter**
```
class RefNulableResultConverter implements JsonConverter<Ref?, dynamic> {
  const RefNulableResultConverter();

  @override
  Ref? fromJson(dynamic json) {
    if (json != null) {
      if (json["@ref"] != null) {
        final refResult = RefResult.fromJson(json["@ref"] as Map<String, dynamic>);
        return refResult.asRef();
      } else if (json["ref"] != null && json["ref"]["collection"] != null) {
        final collection = json["ref"]["collection"] as String;
        final id = json["id"] as String;
        return Ref(Collection(collection), id);
      }
    }
    return null;
  }

  @override
  Ref? toJson(Ref? ref) {
    if (ref?.id == null || ref?.id == "") {
      return null;
    }
    return ref;
  }
}

```

**Example usage**
```
@JsonSerializable()
@RefResultConverter()
@RefNulableResultConverter()
@AlertConverter()
class ShoppingListData {
  Ref accountRef;
  Ref privateUserRef;
  Ref walletRef;
  DateTime createdDate;

  Ref irregularExpensePlanRef;
  String name;
  bool campaignAlert;
  Alert? alert;

  @JsonKey(defaultValue: 0)
  int paymentMethodId = 0;
  Ref? personRef;
  Ref? corporationRef;
  @JsonKey(defaultValue: 1)
  int relevantTypeId = 1;
  Ref? paymentAccountRef;
  Ref? paymentCardRef;

  String? note;
  List<String>? photoUrlList;
  List<String>? audioUrlList;

  ShoppingListData({
    required this.accountRef,
    required this.privateUserRef,
    required this.walletRef,
    required this.createdDate,
    required this.irregularExpensePlanRef,
    required this.name,
    required this.campaignAlert,
    required this.paymentMethodId,
    required this.relevantTypeId,
    this.personRef,
    this.corporationRef,
    this.paymentAccountRef,
    this.paymentCardRef,
    this.alert,
    this.note,
    this.photoUrlList,
    this.audioUrlList,
  });

 

  factory ShoppingListData.fromJson(Map<String, dynamic> json) => _$ShoppingListDataFromJson(json);

  Map<String, dynamic> toJson() => _$ShoppingListDataToJson(this);
}
```

rrousselGit:
You don't need to use both snnotations at the same time anymore. remove RefNulableResultConverter and your code should work

hasimyerlikaya:
Thank you,

Actually, I tried that way, but there is a problem.

PersonRef value should be null, but Ref.empty() value.
There should be a null check during the conversion process and if json['personRef'] is null it should be set to null, no conversion should be done.


image


@rrousselGit

rrousselGit:
It's just a thought but: is changing the `dyanmic`in your JsonConverter to `Object?`fixing the issue?

hasimyerlikaya:
I don't think it will fix it, Ref.empty() value will be assigned because it will still be null.

There should be a null check and if the data is null the converter should not be called at all.

dzluppin:
> Hi,
> 
> I have problem with this merge. I use two converter for a custom type wich name is Ref. I can't generate g.dart file because there is an error.
> 
> "Found more than one matching converter for `Ref?`."
> 
> I don't want to add annotation top of nullable field every time. How can I use both converter with class level annotation?
> 
> **Non-Nullable Converter**
> 
> ```
> class RefResultConverter implements JsonConverter<Ref, dynamic> {
>   const RefResultConverter();
> 
>   @override
>   Ref fromJson(dynamic json) {
>     if (json != null) {
>       if (json["@ref"] != null) {
>         final refResult = RefResult.fromJson(json["@ref"] as Map<String, dynamic>);
>         return refResult.asRef();
>       } else if (json["ref"] != null && json["ref"]["collection"] != null) {
>         final collection = json["ref"]["collection"] as String;
>         final id = json["id"] as String;
>         return Ref(Collection(collection), id);
>       }
>     }
>     return Ref.empty();
>   }
> 
>   @override
>   Ref toJson(Ref ref) {
>     if (ref.id == null || ref.id == "") {
>       return Ref.empty();
>     }
> 
>     return ref;
>   }
> }
> ```
> 
> **Nullable Converter**
> 
> ```
> class RefNulableResultConverter implements JsonConverter<Ref?, dynamic> {
>   const RefNulableResultConverter();
> 
>   @override
>   Ref? fromJson(dynamic json) {
>     if (json != null) {
>       if (json["@ref"] != null) {
>         final refResult = RefResult.fromJson(json["@ref"] as Map<String, dynamic>);
>         return refResult.asRef();
>       } else if (json["ref"] != null && json["ref"]["collection"] != null) {
>         final collection = json["ref"]["collection"] as String;
>         final id = json["id"] as String;
>         return Ref(Collection(collection), id);
>       }
>     }
>     return null;
>   }
> 
>   @override
>   Ref? toJson(Ref? ref) {
>     if (ref?.id == null || ref?.id == "") {
>       return null;
>     }
>     return ref;
>   }
> }
> ```
> 
> **Example usage**
> 
> ```
> @JsonSerializable()
> @RefResultConverter()
> @RefNulableResultConverter()
> @AlertConverter()
> class ShoppingListData {
>   Ref accountRef;
>   Ref privateUserRef;
>   Ref walletRef;
>   DateTime createdDate;
> 
>   Ref irregularExpensePlanRef;
>   String name;
>   bool campaignAlert;
>   Alert? alert;
> 
>   @JsonKey(defaultValue: 0)
>   int paymentMethodId = 0;
>   Ref? personRef;
>   Ref? corporationRef;
>   @JsonKey(defaultValue: 1)
>   int relevantTypeId = 1;
>   Ref? paymentAccountRef;
>   Ref? paymentCardRef;
> 
>   String? note;
>   List<String>? photoUrlList;
>   List<String>? audioUrlList;
> 
>   ShoppingListData({
>     required this.accountRef,
>     required this.privateUserRef,
>     required this.walletRef,
>     required this.createdDate,
>     required this.irregularExpensePlanRef,
>     required this.name,
>     required this.campaignAlert,
>     required this.paymentMethodId,
>     required this.relevantTypeId,
>     this.personRef,
>     this.corporationRef,
>     this.paymentAccountRef,
>     this.paymentCardRef,
>     this.alert,
>     this.note,
>     this.photoUrlList,
>     this.audioUrlList,
>   });
> 
>  
> 
>   factory ShoppingListData.fromJson(Map<String, dynamic> json) => _$ShoppingListDataFromJson(json);
> 
>   Map<String, dynamic> toJson() => _$ShoppingListDataToJson(this);
> }
> ```

I have this kind of issue too.

Some of my classes have both DateTime and DateTime? attributes. And because my server returned a non-standard millisecond format, I had to create a JsonConverter class for it.

Previously, approx. 6 months ago, my code generator worked fine when running build_runner using both the nullable and non-nullable versions of JsonConverter for DateTime, but now it returns an error "Found more than one matching converter for DateTime?."

ljmatan:
For anyone not being able to differentiate between nullable and non-nullable types, the workaround is as below: 

```dart
@JsonSerializable(
  converters: [
    ExampleSerializationHandlerString(),
    ExampleSerializationHandlerBool(),
    ExampleSerializationHandlerInt(),
    ExampleSerializationHandlerDouble(),
    ExampleSerializationHandlerMap(),
  ],
)
@ExampleSerializationHandlerStringNullable()
@ExampleSerializationHandlerBoolNullable()
@ExampleSerializationHandlerIntNullable()
@ExampleSerializationHandlerDoubleNullable()
@ExampleSerializationHandlerMapNullable()
```

This is with the latest json_serializable (6.8.0) and json_annotation (4.9.0) versions at the time of this writing.

PR #1146, opened 2022-05-09
Serializing enums explicitly references types unless nullable (issue …

### LINKED ISSUE #1145: Enum serialization creates nullable type for non-nullable enums
Using an Enum as a key in a map yields a field with type <String?, dynamic> (rather than the expected <String, dynamic>.

```

part 'shape_sorter.g.dart';

enum shapeProperty { size, color }

@JsonSerializable()
class ShapeSorter {
  Map<shapeProperty, String> shapeProperties;

  ShapeSorter(this.shapeProperties);
}
```
This leads to the generated code;
```
Map<String, dynamic> _$ShapeSorterToJson(ShapeSorter instance) =>
    <String, dynamic>{
      'shapeProperties': instance.shapeProperties
          .map((k, e) => MapEntry(_$shapePropertyEnumMap[k], e)),
    };
    

const _$shapePropertyEnumMap = {
  shapeProperty.size: 'size',
  shapeProperty.color: 'color',
};
```

This is because _$shapePropertyEnumMap[k] yields type String?, even though the value - I think - is guaranteed to exist in the map. This isn't a major problem unless you have post processing (I'm using FakeFirebaseFirestore, but there could be otherwise). 

This causes a problem with anything which is doing null-aware runtime typechecking for Map<String, dynamic>.

json_serializable: "6.2.0"

ssabdb:
#1145 I've fixed the issue for my own purposes by inspecting the `targetType` - if not nullable, I use the `!` operator to remove the nullable type outputted by directly indexing `_$EnumTypeEnumMap`.

This includes a unit test which fails prior to the fix.

I considered doing something similar to introducing a helper function used in the enum deserialization (`$enumDecode`), but this just seemed simpler - since the map  `_$EnumTypeEnumMap` is generated from the possible values of the enum, it should never result in a null, unless the variable storing the enum itself is nullable - and if that happens I don't emit an `!` operator.

It does change a number of other generated files though, which might be a concern. 

resolves #1145


kevmoo:
@ssabdb – can you look at the failures?

ssabdb:
Looks like the test builds some example files which are diff'd and have changed to include the null assertion “bang” operator as an intended result of this PR. 

They will need to be regenerated so the test passes - I'll figure it out today or tomorrow.

ssabdb:
@kevmoo this failure looks related to the fix you're working on in #1117?

Let me know if I need to do anything else.

kevmoo:
@ssabdb – the markdown failure is unrelated – I should likely get rid of that.

json_serializable/test/integration/integration_test.dart
```diff
@@ -304,6 +304,14 @@ void main() {
     expect(dayTypeEnumValues, ['no-good', 'rotten', 'very-bad']);
   });
 
+  test('enum serialization has correct type', () {
+    // ignore: omit_local_variable_types
```
kevmoo, line 308: Don't ignore this lint. Drop the local variable type!

json_serializable/test/integration/integration_test.dart
```diff
@@ -304,6 +304,14 @@ void main() {
...
+    final Issue1145Regression cls =
+        Issue1145Regression(status: {Issue1145RegressionEnum.gamma: true});
+    // Not Map<String?, dynamic>
```
kevmoo, line 311: Drop this comment. JSON does not allow null keys (at least I don't think so)

json_serializable/test/integration/integration_test.dart
```diff
@@ -304,6 +304,14 @@ void main() {
...
+    expect(cls.toJson()['status'], const TypeMatcher<Map<String, dynamic>>());
```
kevmoo, line 312: What is this before the change? I'm confused as to the actual new behavior...

kevmoo (changes requested):

json_serializable/test/integration/integration_test.dart
ssabdb: The new behaviour is to return Map<String, dynamic> correctly, hence the typematcher. The previous behaviour was returning Map<String?, dynamic> - which as you say is not permitted by the json spec.

json_serializable/test/integration/integration_test.dart
ssabdb: I'm happy to clarify the comment, but checking that the resultant output is _not_ Map<String**?**, dynamic> is  the intent of the test

json_serializable/test/integration/integration_test.dart
kevmoo: ah! cool!

kevmoo (changes requested):
Please add an entry to the changelog explaining this change, too!

ssabdb:
@kevmoo. Done. I also added a second unit test to test the branching logic was working properly for nullable enums, eg. in a List type.

kevmoo:
Still need a changelog entry then I'll run CI and land! Thanks so much!

ssabdb:
No worries, I'd misunderstood changelog (I thought you were talking about the gitlog!)

I'm on the road at the moment until late next week but will have a shot at getting round to it.

json_serializable/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
   of type `MyClass?`. ([#822](https://github.com/google/json_serializable.dart/issues/822))
 - Added support for `JsonSerializable(converters: <JsonConverter>[])`
   ([#1072](https://github.com/google/json_serializable.dart/issues/1072))
+- Fix issue with serialization of non-nullable enumerations to emit a non-nullable type ([#1146](https://github.com/google/json_serializable.dart/pull/1146))
```
kevmoo, line 7: Wrap this line at 80 columns!

kevmoo (dismissed):

json_serializable/CHANGELOG.md
ssabdb: Done!

Issue #1170, opened 2022-07-08
Code generation for complex types with default value produces warning


guenth39:
When having a constructor with a not nullable parameter that is a complex type and has a default value, flutter gives me the following warning:
```
: Warning: Operand of null-aware operation '??' has type 'Duration' which excludes null.
lib/model/my_class.g.dart:13
 - 'Duration' is from 'dart:core'.
      d: Duration(microseconds: json['d'] as int) ?? Duration.zero,
```

See the following example code to reproduce it:
```dart
part 'my_class.g.dart';

@JsonSerializable()
class MyClass {
  const MyClass({
    required this.a,
    required this.c,
    this.b = '',
    this.d = Duration.zero,
  });

  factory MyClass.fromJson(Map<String, dynamic> json) =>
      _$MyClassFromJson(json);

  final String a, b;
  final Duration c, d;

  Map<String, dynamic> toJson() => _$MyClassToJson(this);
}
```
``` dart
// GENERATED CODE - DO NOT MODIFY BY HAND

part of 'my_class.dart';

// **************************************************************************
// JsonSerializableGenerator
// **************************************************************************

MyClass _$MyClassFromJson(Map<String, dynamic> json) => MyClass(
      a: json['a'] as String,
      c: Duration(microseconds: json['c'] as int),
      b: json['b'] as String? ?? '',
      d: Duration(microseconds: json['d'] as int) ?? Duration.zero,
    );

Map<String, dynamic> _$MyClassToJson(MyClass instance) => <String, dynamic>{
      'a': instance.a,
      'b': instance.b,
      'c': instance.c.inMicroseconds,
      'd': instance.d.inMicroseconds,
    };

```

**flutter --version**
```
Flutter 3.0.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision cd41fdd495 (vor 4 Wochen) • 2022-06-08 09:52:13 -0700
Engine • revision f15f824b57
Tools • Dart 2.17.3 • DevTools 2.12.2
```

kevmoo:
Fix is on the way!

kevmoo:
published as 6.3.1

guenth39:
Thank u for the fast fix! 🚀 👏

AlexeyKatsuro:
Hello @kevmoo !

Is it possible to adjust codegen behaviour for Duration class and replace `inMicroseconds` with `inSeconds` and `Duration(microseconds: ),` with `Duration(seconds: ),` for whole project?

[...]

PR #1172, opened 2022-07-08
Fix support for Duration fields with default values

kevmoo:
Fixes https://github.com/google/json_serializable.dart/issues/1170

### DIFF
--- a/json_serializable/lib/src/type_helpers/to_from_string.dart
+++ b/json_serializable/lib/src/type_helpers/to_from_string.dart
@@ -5,7 +5,7 @@
 import 'package:analyzer/dart/element/type.dart';
 import 'package:source_gen/source_gen.dart';
 
-import '../type_helper.dart';
+import '../utils.dart';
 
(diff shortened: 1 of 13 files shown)


json_serializable/lib/src/type_helpers/to_from_string.dart
```diff
@@ -76,7 +76,8 @@ class ToFromStringHelper {
 
     final parseParam = isString ? expression : '$expression as String';
 
-    return commonNullPrefix(nullable, expression, '$_parse($parseParam)')
-        .toString();
+    final output = '$_parse($parseParam)';
+
+    return nullable ? ifNullOrElse(expression, 'null', output) : output;
```
kevmoo, line 81: This was the only use of `commonNullPrefix` – just inlined it here!

natebosch:
The commit message should say what was broken and how this fixes it. Can you edit the first comment with this info?

natebosch (approved):

Issue #891, opened 2021-05-12
final fields are not written during serialization


MatrixDev:
### Source code
```dart
@JsonSerializable
class MyClass {
  @JsonKey(name: 'value1')
  var value1 = '';

  @JsonKey(name: 'value2')
  final value2 = 'must always be this';
}
```

### Generated code:
```dart
MyClass _$MyClassFromJson(Map<String, dynamic> json) {
  return MyClass()..value1 = json['value1'] as String;
}

Map<String, dynamic> _$MyClassToJson(MyClass instance) => <String, dynamic>{
      'value1': instance.value1,
    };
```

### Actual result
vars are written, finals are not

### Expected result
all fields must be written - vars and finals

### Versions
```
Flutter 2.0.6
json_annotation: ^4.0.0
json_serializable: ^4.1.0
```

kevmoo:
This is by design. We only "round-trip" values that can be provided when we create the value.

You can always wrap the generated value and add in the final value.

See https://github.com/google/json_serializable.dart/issues/274

MatrixDev:
@kevmoo is there any reasoning for this? it is very strange that **explicitly** saying that I want to serialize something does nothing. At very least codegen should throw an error if this is the case. Explicit stuff cannot be simply ignored, it took me some time to find why API calls where not working because of this.

esDotDev:
This is bizarre... we have a type, we need to write it into our payload so the server knows what class this is. 

I don't understand the rationale.

```
class Address {
  final String street;

  const Address({
    this.street = '',
    this.street2 = '',
    this.city = '',
    this.state = '',
    this.postalCode = '',
    this.country = '',
  });

  final String type= "Address"; // We need this in the payload... 
}
```

kevmoo:
I'll reopen this issue. PR's are welcome!

sgehrman:
This sounds crazy.  Are you saying this won't work:

I haven't tried json_serializable yet, I'm looking for a replacement for simple_json (which is the best, but not being updated)

I can't have finals?

class AnnotationModel {
  AnnotationModel({
    required this.containerId,
    required this.content,
    required this.createdAt,
    required this.id,
    required this.modifiedAt,
    required this.status,
  });

  final String containerId;
  final String content;
  final DateTime createdAt;
  final String id;
  final DateTime modifiedAt;
  final String status;
}

Issue #24, opened 2017-07-23
Looking into setter-only or getter-only


kevmoo:
* I guess you could have a setter-only for just `fromJson` case.
* Don't try to create a write to a property that's only a setter
* etc...

edlman:
Hello, I'm not sure wheter to open new issue or just comment this one. I have a model on server side where I have two getters which return values base on other properties and I want them called from _$ModelToJson(). I have no setters (no reason to have them).
```
Model {
  @JsonKey(name: 'holder')
  String holder;
  @JsonKey(name: 'source')
  String source;
  @JsonKey(name: 'symbol')
  String get symbol => holder == 'owner' ? 'O' : 'G';
  @JsonKey(name: 'description')
  String get description => holder == 'owner' ? 'Owner' : 'Guest';
}
```
and I expect _$ModelToJson to contain
```
Map<String, dynamic> _$ModelToJson(Model instance) => <String, dynamic>{
      'holder': instance.holder,
      'source': instance.source,
      'symbol': instance.symbol,
      'description': instance.description
}
```
but `symbol` and `description` are not included until I create setters, so then setters are called from `$_ModelFromJson(Map<String, dynamic> json) { ... }`. 
Simple workaround is to create setters which do nothing, but I suppose that's not correct.
Vice versa I would expect that when I only have setters the `_$ModelFromJson()` should call them and `_$ModelToJson()` should ignore these properties.

kevmoo:
@edlman do you need to create this class from Json?

If not, just put `@JsonSerializable(createFactory: false)` on the class and you should be good!

edlman:
@kevmoo, yes I need to create this class from json (or other map, e.g. database Row). It's on the server side, some data are fetched from PgSQL to the object using _$ModelFromJson(PgRow as Map) and then passed as JSON to a client using _$ModelToJson(). And I need to pass some server generated values from getters to the client.

edlman:
Practically yes. When I use getters I expect to have simplier JSON on input to _$ModelFromJson (either from DB or from client). On output from $_ModelToJson I expect richer JSON containing values not only from variables but also from getters.
And vice versa when I use setters I expect richer JSON on input to $_ModelFromJson, I can process these values in setters for whatever I need. On output from _$ModelFromJson I expect simplier JSON contaning only values from variables.
As I explained on my example code, I have some values generated by getters on server side. On client side I process these values from JSON (and I would like to process them with setters eventually).
[...]

PR #1178, opened 2022-07-15
More control over which fields should be added to fromJson and toJson

joeldomke:
Adds the option to `JsonKey` to include fields in `toJson` that are not used in the factory (`fromJson`).

Related issues: #24, #891


joeldomke:
I guess [this approach ](https://github.com/google/json_serializable.dart/issues/569#issuecomment-953354620) you suggested could also be used. Let me know if you want me to take a look at that.

joeldomke:
Okay, I should be able to implement this. Two questions:

1. Currently there exists the `unavailableReason` 'Setter-only properties are not supported'. I guess this should no longer be the case, since we offer `IncludeWith.fromJson`? But this would also be a breaking change for classes that have fields that only have a setter, which would then have to be annotated with `@JsonKey(includeWith: IncludeWith.ignore)`.
2. Same for getters. None of the 4 `IncludeWith` options match the old behaviour where getters without setters would be ignored. This would mean that all those getters would have to be annotated with `@JsonKey(includeWith: IncludeWith.ignore)`. 

Should there be a fifth option, that matches the old behaviour, to prevent those breaking changes?

json_serializable/test/src/include_with_test_input.dart
```diff
@@ -0,0 +1,90 @@
...
+''')
+@JsonSerializable()
+class IncludeWithFromJson<T extends num, S> {
+  @JsonKey(includeWith: IncludeWith.fromJson)
+  late String someString;
+
+  @JsonKey(includeWith: IncludeWith.fromJson)
+  String get setter => someString;
+
+  set setter(String value) => someString = value;
```
joeldomke, line 87: Since the `JsonKey` annotation can't be added to setters, we still need a getter.

json_annotation/lib/src/json_key.dart
```diff
@@ -136,3 +136,5 @@ class JsonKey {
 }
 
 enum _NullAsDefault { value }
+
+enum IncludeWith { both, toJson, fromJson, ignore, legacy }
```
joeldomke, line 140: I added a fifth option, currently called `legacy`. Fields that use this option should behave as before, so currently this is the default option.

json_serializable/test/src/include_with_test_input.dart
joeldomke: I guess it would be possible to also support fields that only have setters? @kevmoo would you consider this PR, if I implemented this, or are there other issues with this PR or just limited time/resources from your side?

kevmoo:
So. I think I like this BUT! it's a breaking change.

We *could* bring it in as a non-breaking change, but we'd need to keep the existing `ignore` field.

Are you on Twitter ? Can you DM me so we can schedule time to chat about this?

joeldomke:
I just saw that you enabled the CI. This PR was a bit of a mess. Two different approaches, only code and no explanation. I am currently preparing a new PR that addresses the `ignore` related breaking change. It will have a more detailed description and a list of open questions. So you don't have to waste any time on this. I should be able to get it out next week.

kevmoo:
Can't run with this. Need to have a model that support the old and the new living together.

PR #1193, opened 2022-08-28
Fix JsonConverter docs example

### LINKED ISSUE #1194: Fix JsonConverter docs example
The docs have an example where the Converter class name differs from when it's used as an annotation.

feinstein:
The docs have an example where the Converter class name differs from when it's used as an annotation.

Closes #1194.


kevmoo:
Thanks, @feinstein !

kevmoo (approved):

Issue #1147, opened 2022-05-10
[Feature Request] Enum value with enhanced Enums


a7md0:
Feature request related to [Enhanced enums](https://github.com/dart-lang/language/blob/master/accepted/future-releases/enhanced-enums/feature-specification.md) while using [JsonValue](https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonValue-class.html)

1. Currently, we would have to declare the value twice per Enum value (to allow this package to serialize/deserialize the Enum and to have access to `.value`)
```dart
import 'package:json_annotation/json_annotation.dart';

part 'status_code.g.dart';

enum StatusCode {
  @JsonValue(200)
  success(200),

  @JsonValue(500)
  weird(500),

  const StatusCode(this.value);

  final int value;
}
```

It would be much better if we could have the generated map `EnumMap` to use the value specified in the constructor

a7md0:
I have no idea how could generation work in the package, but one possibility will be offer some kind of interface `EnumValue<T>` to be implemented by the Enum, thus making the `T value` required to be implemented. Next it would be to check for Enums implementing this interface and use the value passed in the constructor to create the `EnumMap` with the code generation.

kevmoo:
Or we could "steal" the `JasonValue` annotation and put it on the `enum` definition with the field name – in the case above `"value"`.

The package is not setup to support adding language-specific features until we're ready to bump the min SDK, so it'll likely be a while until we get there.

kevmoo:
https://github.com/google/json_serializable.dart/pull/1180 bumps the min version of the required SDK – we could entertain this now!

memoriasIT:
Hey, did this get any updates? I am running the latest version and it's still generating by name. Am I missing something?

kevmoo:
This feature has been added as of

https://pub.dev/packages/json_annotation/versions/4.7.0
and
https://pub.dev/packages/json_serializable/versions/6.4.0

YiannisBourkelis:
@kevmoo  Sorry for coming back on this. Just to make things clear, when using the new syntax of enhanced enums should we still use the @JsonValue annotation not top of each enum value or does the library takes care of it?

Looks that json_convert always generates a string representation of the enum name when there is no @JsonValue.

kevmoo:
Just use `@JsonEnum` on the enum itself. `@JsonValue` should no longer be needed.

YiannisBourkelis:
@kevmoo Perfect! Thank you

PR #1203, opened 2022-09-20
Add JsonEnum.valueField for encoding enhanced enum values

kevmoo:
Fixes https://github.com/google/json_serializable.dart/issues/1147

Prepare to release json_annotation v4.7.0


json_serializable/lib/src/enum_utils.dart
```diff
@@ -33,20 +33,50 @@ String? enumValueMapFromType(
...
+        } else {
+          throw InvalidGenerationSourceError(
+            '`JsonEnum.valueField` was set to "${jsonEnum.valueField}", but '
+            'that field does not have a type of String, int, or null.',
+            element: targetType.element2,
+          );
+        }
+      } else {
+        fieldValue = encodedFieldName(jsonEnum.fieldRename, fe.name);
+      }
```
natebosch, line 65: This method was 37 lines, it's now 61 lines. It previously had 2 levels of nesting conditionals, now it has 3 levels.

Consider refactoring to split into smaller chunks.

json_serializable/test/integration/json_enum_example.dart
```diff
@@ -25,6 +25,19 @@ enum DayType {
 
 Iterable<String> get dayTypeEnumValues => _$DayTypeEnumMap.values;
 
+@JsonEnum(alwaysCreate: true, valueField: 'value')
```
natebosch, line 28: [probably out of scope] If we changed this to use `Symbol` then it would be limited to legal (but not necessarily correct) names for free. I don't know how many `String` arguments we currently use that correspond to Dart names. If this is the first we can consider starting with `Symbol`. If there are bunch of others we might use `String` here for consistency.

```dart
@JsonEnum(alwaysCreate: true, valueField: #value)
```

natebosch (approved):

json_serializable/lib/src/enum_utils.dart
kevmoo: https://github.com/google/json_serializable.dart/pull/1204

json_serializable/test/integration/json_enum_example.dart
kevmoo: Not out of scope! I like it!

kevmoo: Nope!
https://api.dart.dev/stable/dart-core/Symbol-class.html
No way to (cleanly) get the string value out of the symbol. I'd have to do a regexp through `Symbol(fieldName)` – which feels hacky.

Also would need to teach folks about `Symbol` syntax.

natebosch: > No way to (cleanly) get the string value out of the symbol.

https://pub.dev/documentation/analyzer/latest/dart_constant_value/DartObject/toSymbolValue.html

kevmoo: Yeah, but then I can't use `JsonEnum` in the generator. I need to create ANOTHER class that has a string property instead of a `Symbol` property.

I don't think it's worth it. Also not consistent w/ the other fields in the package. 🤷

PR #1204, opened 2022-09-20
json_serializable: refactor enum logic

kevmoo:
Refactor a function into to functions
Simplify the return value

### DIFF
--- a/json_serializable/lib/src/enum_utils.dart
+++ b/json_serializable/lib/src/enum_utils.dart
@@ -27,66 +27,14 @@ String? enumValueMapFromType(
     return null;
   }
 
-  MapEntry<FieldElement, dynamic> generateEntry(FieldElement fe) {
-    final annotation =
-        const TypeChecker.fromRuntime(JsonValue).firstAnnotationOfExact(fe);
-
-    dynamic fieldValue;
-    if (annotation == null) {
-      if (jsonEnum.valueField != null) {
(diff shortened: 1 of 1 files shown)


kevmoo:
FYI @natebosch – you might could disable auto-merge if you think I may want to change things given your feedback.

json_serializable/lib/src/enum_utils.dart
```diff
@@ -27,73 +27,78 @@ String? enumValueMapFromType(
...
+
+  final items = enumMap.entries
+      .map((e) => '  ${targetType.element2!.name}.${e.key.name}: '
+          '${jsonLiteralAsDart(e.value)},')
+      .join();
+
+  return 'const ${constMapName(targetType)} = {\n$items\n};';
+}
+
+Object? generateEntry({
```
natebosch, line 47: [nit] can this be private?

natebosch (approved):

json_serializable/lib/src/enum_utils.dart
natebosch: Ugh, I need to remember to check the "request changes" box when I leave comments...

natebosch:
> FYI @natebosch – you might could disable auto-merge if you think I may want to change things given your feedback.

Yeah, I thought about that after the last one got merged and then didn't think about it at all when reviewing this one...

I'll get the reflexes to check that box at some point 😝

kevmoo:
Ha!

PR #1221, opened 2022-10-08
Custom Map Types

TimWhiting:
@kevmoo
Iterated as requested on #997 

Fixes: https://github.com/google/json_serializable.dart/issues/396

Custom map types with fromJson / toJson methods can now have keys automatically de/serialized to strings. 

Usage:

Custom map types can make the fromJsonK / toJsonK parameters of their fromJson / toJson methods detected as map keys by using String? rather than Object? as the serialized type.

```dart
// Custom map type with custom serializers
class CustomMap<K, V> {
  final Map<K, V> map;

  CustomMap(this.map);

  factory CustomMap.fromJson(
    Map<String, dynamic> json,
    K Function(String?) fromJsonK,
    V Function(Object?) fromJsonV,
  ) =>
      CustomMap(json.map<K, V>(
          (key, value) => MapEntry(fromJsonK(key), fromJsonV(value))));

  Map<String?, dynamic> toJson( 
          String? Function(K) toJsonK,  // Json serializable detects this as a custom Key value because of the (String?).
 Object? Function(V) toJsonV) =>
      map.map((key, value) => MapEntry(toJsonK(key), toJsonV(value)));
}

// User of custom map type & JsonSerializable
@JsonSerializable()
class UseOfCustomMap {
  final CustomMap<int, String> map;

  UseOfCustomMap(this.map);

  factory UseOfCustomMap.fromJson(Map<String, dynamic> json) =>
      _$UseOfCustomMapFromJson(json);

  Map<String, dynamic> toJson() => _$UseOfCustomMapToJson(this);
}
```

### DIFF
--- a/json_serializable/CHANGELOG.md
+++ b/json_serializable/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.6.0
+- Allow custom map types to automatically have key types serialized. 
+  Fixes[#396](https://github.com/google/json_serializable.dart/issues/396)
+
 ## 6.5.0
 
 - Allow constructors to be passed to `JsonKey` parameters that support
--- a/json_serializable/lib/src/type_helpers/json_helper.dart
+++ b/json_serializable/lib/src/type_helpers/json_helper.dart
@@ -12,9 +12,11 @@ import 'package:source_helper/source_helper.dart';
 
 import '../default_container.dart';
 import '../type_helper.dart';
+import '../unsupported_type_error.dart';
 import '../utils.dart';
 import 'config_types.dart';
 import 'generic_factory_helper.dart';
+import 'map_helper.dart';
 
 const _helperLambdaParam = 'value';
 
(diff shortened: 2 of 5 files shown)


kevmoo:
Hrm...I'm REALLY worried about the complexity added here – for a pretty narrow scenario.

TimWhiting:
> Hrm...I'm REALLY worried about the complexity added here – for a pretty narrow scenario.

The complexity of auto-detection of when this applies, or the complexity of the implementation?

I actually thought that the implementation and auto-detection were relatively straightforward. Plumbing through a flag from the annotation might be less 'magic', but would require far more changes.

As far as it being a narrow scenario.
The goal of json_serializable in my opinion is to automatically generate serialization methods for data models.
If this feature was trying to serialize objects that aren't data models, I would see the value in your argument.

I would argue that using immutable collections in your data model is a fairly common practice, and should be a use case we support.

JoanSchi:
@TimWhiting 

DateTime, BigInt url for example as key works  fine without hacky workaround, because these types are implemented correctly, double int, bool are not parsed correctly I think that works for the hacky woraround.
See issue #1332. and issue: https://github.com/marcglasberg/fast_immutable_collections/issues/58. 

With DateTime I get a cast error because the DateTime is parsed correctly with  toJsonK and then _safeKeyToJson tries to parse the already converted String into DateTime again in a DatateTime. -> error.

Maybe it is possible to use only hacky workaround for int double bool, until  #1332 is fixed. :)

Issue #1249, opened 2022-11-16
`JsonEnum` allow to use `index` with `valueField`


Zekfad:
Currently you can't set `valueFiled` in `JsonEnum` to `index`:

```log
`JsonEnum.valueField` was set to "index", but that is not a valid, instance field on `MyEnum?`.
```

As part of enhanced enums, every enum has `index` property (and `name`, actually, since 2.15).
So it would be helpful, to avoid making field that replicates `index` field.

* Flutter 3.6.0-4.0.pre.8 • channel master • https://github.com/flutter/flutter.git
* Framework • revision a2233eabed (2 hours ago) • 2022-11-16 10:20:45 -0800        
* Engine • revision 223e1f092d
* Tools • Dart 2.19.0 (build 2.19.0-406.0.dev) • DevTools 2.19.0

kevmoo:
Good idea!

PR #1250, opened 2022-11-17
Support `JsonEnum.valueField` being set with `'index'`

kevmoo:
Fixes https://github.com/google/json_serializable.dart/issues/1249

### DIFF
--- a/json_serializable/lib/src/enum_utils.dart
+++ b/json_serializable/lib/src/enum_utils.dart
@@ -88,6 +88,20 @@ Object? _generateEntry({
 
       final e = fieldElementType.getField(valueField);
 
+      if (e == null && valueField == 'index') {
+        final values = fieldElementType.getField('values')!;
+
+        final items = ConstantReader(values.computeConstantValue()).listValue;
+
+        return items
+            .singleWhere((element) {
+              final name = element.getField('_name')!.toStringValue()!;
+              return name == field.name;
+            })
+            .getField('index')!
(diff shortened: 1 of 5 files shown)


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

json_serializable/lib/src/enum_utils.dart
```diff
@@ -88,6 +88,20 @@ Object? _generateEntry({
...
       final e = fieldElementType.getField(valueField);
 
+      if (e == null && valueField == 'index') {
+        final values = fieldElementType.getField('values')!;
+
+        final items = ConstantReader(values.computeConstantValue()).listValue;
+
+        return items
+            .singleWhere((element) {
+              final name = element.getField('_name')!.toStringValue()!;
```
jakemac53, line 98: is `_name` here something that is guaranteed to stay the same?

scheglov: I'm not sure what you are trying to achieve here. You have `FieldElement field` which is a constant in an `EnumElement`. Then you try to find the corresponding `DartObject` value in its fields, using the name, and get back its index in `values`. If I understand it correctly, then why not use `enumElement.fields.where((field) => field.isEnumConstant).toList().indexOf(fieldElement)`?

kevmoo: Thanks! done!

jakemac53 (approved):

PR #1274, opened 2023-01-31
Enable and fix new lints

kevmoo:


_test_yaml/test/ensure_build_test.dart
```diff
@@ -4,6 +4,8 @@
 
 @TestOn('vm')
 @Tags(['presubmit-only'])
+library test;
```
kevmoo, line 7: To be cleaned up when we have https://github.com/dart-lang/linter/issues/4031

jakemac53, line 7: what is up with these?

kevmoo: We don't want dangling annotations. These are annotating the LIBRARY.

jakemac53: oh I see, it does make things more stable I guess so sgtm

jakemac53 (approved):

Issue #1222, opened 2022-10-08
Support for Records


TimWhiting:
With the upcoming records feature, it would be good to figure out how to serialize those with this package.

I'd propose:
- Make it an error to mix positional and named fields in an **inline type**
- Allow records with only named fields - serializes to a normal JSON map with keys as the field names
- Allow records with only positional fields - serializes to a normal JSON list (note that the list might not be homogenous)

Allow records with mixed fields with the following qualifications:
- only via a typedef - so that we don't have to canonicalize structural types.
- is converted to a JSON map
- positional fields require a JsonKey annotation with a name

Potential other requirements to consider:
- Only support JsonKey annotation with typedefs - to make it clear we don't canonicalize structural types

Note that we could probably support not requiring to go through a typedef, it just makes it more error prone for users, because they need to understand that the type is not canonicalized and the annotation would have to be added to all records.

Finally, we would need to consider what annotating a record type by itself (not embedded in an object) with JsonSerializable would look like. I think creating a toJson extension method would be a good approach. Where to put the fromJson method / how to name it is a bit more tricky.

kevmoo:
I'm not thinking about this too much (yet) – but a good thing to discuss early!

shilangyu:
Probably worth reviving now that records reached stable?

humbertoadan:
> Probably worth reviving now that records reached stable?

Yes, it's definitely needed now. I'm actually trying JsonKey.toJson and JsonKey.fromJson to use custom conversion on an annotated field which is a list of records. If I do this: List<({String id, int quantity})>? records, it gives me a "Null check operator used on a null value" error.

kevmoo:
Hrm...perhaps...

I'm a bit nervous about the effort required here. I could dig a bit.

eximius313:
I've just upgraded to Dart 3.0 and encountered this problem while generating mocks (using Mockito):
```
[SEVERE] json_serializable on lib/src/my_class.dart (cached):

This builder requires Dart inputs without syntax errors.
However, package:my_pkg/src/my_class.dart (or an existing part) contains the following errors.
my_class.dart:56:10: This requires the 'records' language feature to be enabled.

Try fixing the errors and re-running the build.
```

I've set 
```
analyzer:
  enable-experiment:
    - records
```
in `analysis_options.yaml` but this didn't help

kevmoo:
You need to get the latest `pkg:analyzer`, too!

eximius313:
[...]

PR #1314, opened 2023-05-11
Support Dart 3 Record types

kevmoo:
Closes https://github.com/google/json_serializable.dart/issues/1222

### DIFF
--- a/json_serializable/lib/src/type_helpers/record_helper.dart
+++ b/json_serializable/lib/src/type_helpers/record_helper.dart
@@ -1,0 +1,124 @@
+import 'package:analyzer/dart/element/type.dart';
+import 'package:source_helper/source_helper.dart';
+
+import '../type_helper.dart';
+import '../utils.dart';
+
+class RecordHelper extends TypeHelper<TypeHelperContextWithConfig> {
+  const RecordHelper();
+
+  @override
+  Object? deserialize(
+    DartType targetType,
+    String expression,
+    TypeHelperContextWithConfig context,
+    bool defaultProvided,
+  ) {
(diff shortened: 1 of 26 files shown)


kevmoo:
The implementation is less than 200 lines. Most of this is auto-generated test code.

json_serializable/lib/src/type_helpers/record_helper.dart
```diff
@@ -0,0 +1,124 @@
...
+    bool defaultProvided,
+  ) {
+    if (targetType is! RecordType) return null;
+
+    final items = <Object>[];
+
+    const paramName = r'$jsonValue';
+
+    var index = 1;
+    for (var field in targetType.positionalFields) {
```
jakemac53, line 24: It feels a bit weird to serialize the positional fields as a map with the `$*` names. Although it does simplify things some.

I could also see serializing them as a List, with positional fields followed by a nested Map for the named fields, so like (1, 2, {name: 'Foo'}) would be like `[1, 2, {"name": "Foo"}]`.

Ultimately I don't know how much it matters how we serialize records, but that would be a bit more true to the original format?

jakemac53 (approved):

json_serializable/lib/src/type_helpers/record_helper.dart
kevmoo: I don't think we gain anything by encoding to look like the definition syntax. It makes encode and decode more complex. I'm trying to do the thing that's just easiest.

Dampfwalze:
Are there examples or other documentation on how to serialize and deserialize records?

PR #1323, opened 2023-05-20
[Annotation Only] Adding case insensitive to enum based on loading JsonEnum config

matanshukry:
Part of https://github.com/google/json_serializable.dart/pull/1155 - see details here.

This is just the annotation part of the PR.


json_annotation/lib/src/json_key.dart
```diff
@@ -149,19 +149,26 @@ class JsonKey {
...
+        bool? nullable,
     this.defaultValue,
     this.disallowNullValue,
     this.fromJson,
     @Deprecated(
       'Use `includeFromJson` and `includeToJson` with a value of `false` '
       'instead.',
     )
-    this.ignore,
+        this.ignore,
```
matanshukry, line 171: fyi this was done by dart format 🤷

json_annotation/lib/src/json_key.dart
```diff
@@ -149,19 +149,26 @@ class JsonKey {
   /// valid on a nullable enum field.
   final Enum? unknownEnumValue;
 
+  /// If true, enum will be parsed with case-insensitive.
+  /// Specifically, both values will be lower-cased and compared.
+  ///
+  /// Valid only on enum fields with a compatible enum value.
+  final bool caseInsensitive;
```
t-beckmann, line 156: Why is the option needed on both `JsonEnum` and `JsonKey`?

matanshukry: @t-beckmann so it can be used both at the enum level and at the field level.

Your question did make me add tests to the original PR, so I also found an issue and fixed it, and added 3 test cases: case insensitive on enum level, case insensitive on field level, and without any case insensitive (verifying behavior is same without it).

Example from the other PR:
* On the enum: https://github.com/google/json_serializable.dart/pull/1155/files#diff-1ac6a89c51c26fa87981cde2ede9cfa268021350a1bdc2770e982feea1dcf242R9
* On the field: https://github.com/google/json_serializable.dart/pull/1155/files#diff-976667989c3ddcc9bb6d6470103769f5aa9c376f142a64c76fadb4b52f03b736R58

json_annotation/lib/src/json_key.dart
kevmoo: You need to be using a dev release of Dart 3.1 for this.

PR #1329, opened 2023-06-14
Unify dart and flutter generation instructions

JAicewizard:
When using `flutter pub run ...` it prints "Deprecated. Use `dart run` instead." (flutter 3.10) so this removes the instructions for flutter and just instructions and just use `dart run ...`

### DIFF
--- a/json_serializable/README.md
+++ b/json_serializable/README.md
@@ -75,10 +75,7 @@ Map<String, dynamic> _$PersonToJson(Person instance) => <String, dynamic>{
 Once you have added the annotations to your code you then need to run the code
 generator to generate the missing `.g.dart` generated dart files.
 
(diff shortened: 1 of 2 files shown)


kevmoo (approved):
LOVE IT! Thank you!

kevmoo:
@JAicewizard – FYI – the readme is generated from a template. I pushed the fix here. It's weird.

JAicewizard:
Ah yeah sure, thanks for fixing that I had no idea :)

kevmoo:
> Ah yeah sure, thanks for fixing that I had no idea :)

I'm kinda proud of the setup to make sure a bunch of stuff says in sync – but it does increase the contributor burden. It's a VERY one-off solution!

PR #1335, opened 2023-07-04
fix: update pub_semver and build_runner version

ipcjs:
Older versions of `build_runner` and `pub_semver` report the following error when executing `dart run build_runner`. So need to update them.

```
Building package executable... (4.1s)
Failed to build build_runner:build_runner:
/Users/ipcjs/.pub-cache/hosted/pub.dev/build_runner-2.3.2/lib/src/build_script_generate/bootstrap.dart:76:40: Error: Method not found: 'NullThrownError'.
      final error = e[0] as Object? ?? NullThrownError();
                                       ^^^^^^^^^^^^^^^
/Users/ipcjs/.pub-cache/hosted/pub.dev/pub_semver-2.1.1/lib/src/version_constraint.dart:96:13: Error: Method not found: 'FallThroughError'.
      throw FallThroughError();
            ^^^^^^^^^^^^^^^^
```

### DIFF
--- a/json_serializable/pubspec.yaml
+++ b/json_serializable/pubspec.yaml
@@ -24,7 +24,7 @@ dependencies:
   json_annotation: '>=4.8.1 <4.9.0'
   meta: ^1.3.0
   path: ^1.8.0
-  pub_semver: ^2.0.0
+  pub_semver: ^2.1.3
   pubspec_parse: ^1.0.0
   source_gen: ^1.3.2
   source_helper: ^1.3.0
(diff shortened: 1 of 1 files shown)


kevmoo:
I get the idea here. Generally we DON'T do updates that just update lower bounds. Folks would have to upgrade to the latest version of this package to get these constraint bumps! (Does that make sense?)

ipcjs:
Updating lower bounds can help developer deal with the problem of incorrect dependency versions.

It actually took me an hour to find out that the reason my code was not generating was because the version of `pub_semver` and `build_runner` was too low.🤔️

CicadaCinema:
In case you are interested, there is an ongoing effort at BMW that can help to address this class of bugs: https://github.com/bmw-tech/dart_apitool

Incorrect lower bound constraints were also what I worked on as part of GSOC last summer and you can read about my findings here: https://arseny.uk/posts/gsoc2022/

PR #1343, opened 2023-08-06
select appropriate converter for nullable fieldtype

### LINKED ISSUE #1339: Found more than one matching converter for `int?`.
my json convertor 
```class JsonIntConverter extends JsonConverter<int, dynamic> {
  const JsonIntConverter();

  @override
  int fromJson(dynamic json) {
    if (json is int) {
      return json;
    } else {
      return int.parse(json);
    }
  }

  @override
  dynamic toJson(int object) {
    return object;
  }
}

class JsonNullIntConverter extends JsonConverter<int?, dynamic> {
  const JsonNullIntConverter();

  @override
  int? fromJson(dynamic json) {
    if (json == null) {
      return null;
    }
    if (json is int) {
      return json;
    } else {
      return int.parse(json);
    }
  }

  @override
  dynamic toJson(int? object) {
    return object;
  }
}
```
my entity
```
@JsonSerializable(converters: [const JsonNullIntConverter,const JsonIntConverter()])
class FileInfo {
  @JsonKey(name: 'status')
  int status;

  @JsonKey(name: 'file_id')
  int fileId = 0;

  @JsonKey(name: 'file_name')
  String? fileName;

  @JsonKey(name: 'file_path')
  String? filePath;

  @JsonKey(name: 'file_preview_url')
  String? filePreviewUrl;

  @JsonKey(name: 'file_size')
  int? fileSize = 0;
}
```

error
```Found more than one matching converter for `int?`.```

if i remove JsonIntConverter,JsonNullIntConverter di not gen code for int
```
FileInfo _$FileInfoFromJson(Map<String, dynamic> json) => FileInfo(
      status: json['status'] as int? ?? 0,
      fileId: json['file_id'] as int? ?? 0,
      fileName: json['file_name'] as String?,
      filePath: json['file_path'] as String?,
      filePreviewUrl: json['file_preview_url'] as String?,
      fileSize: json['file_size'] as int? ?? 0,
      bucket: json['bucket'] as String?,
      key: json['key'] as String?,
      host: json['host'] as String?,
      accessid: json['accessid'] as String?,
      callback: json['callback'] as String?,
      callbackVar: json['callback-var'] as String?,
      policy: json['policy'] as String?,
      signature: json['signature'] as String?,
      expire: json['expire'] as String?,
    );
```

### PR DESCRIPTION
**Pull request by julek-kal**
## Connection with issue(s)

Close #1339 

## Solution description
When `targetType` is nullable and the class is annotated with both converters for nullable and non-nullable type, `_compatibleMatch` returns two of them. So I've added a check if `targetType` is nullable and `matchingAnnotations` length is equal to 2, if so I proceed to try to find a non-null converter and remove it from the list of `matchedConverters`. 

## To Do

- [x] Check the original issue to confirm it is fully satisfied
- [x] Add solution description to help guide reviewers
- [x] Add unit test to verify new or fixed behaviour
- [ ] If apply, add documentation to code properties and package readme

### DIFF
--- a/json_serializable/lib/src/type_helpers/json_converter_helper.dart
+++ b/json_serializable/lib/src/type_helpers/json_converter_helper.dart
@@ -202,6 +202,14 @@ _JsonConvertData? _typeConverterFrom(
     return null;
   }
 
+  if (targetType.isNullableType && matchingAnnotations.length == 2) {
+    final indexOfNonNullableConverter = matchingAnnotations
+        .indexWhere((element) => !element.fieldType.isNullableType);
+    if (indexOfNonNullableConverter >= 0) {
+      matchingAnnotations.removeAt(indexOfNonNullableConverter);
+    }
+  }
+
   if (matchingAnnotations.length > 1) {
     final targetTypeCode = typeToCode(targetType);
     throw InvalidGenerationSourceError(
--- a/json_serializable/test/json_serializable_test.dart
+++ b/json_serializable/test/json_serializable_test.dart
@@ -91,6 +91,7 @@ const _expectedAnnotatedTests = {
   'JsonConvertOnField',
   'JsonConverterCtorParams',
   'JsonConverterDuplicateAnnotations',
+  'JsonConverterIssue1339',
   'JsonConverterNamedCtor',
   'JsonConverterNullableToNonNullable',
   'JsonConverterOnGetter',
--- a/json_serializable/test/src/json_converter_test_input.dart
+++ b/json_serializable/test/src/json_converter_test_input.dart
@@ -126,6 +126,38 @@ class JsonConverterDuplicateAnnotations {
   late Duration value;
 }
 
+@ShouldGenerate(r'''
+JsonConverterIssue1339 _$JsonConverterIssue1339FromJson(
+        Map<String, dynamic> json) =>
+    JsonConverterIssue1339()
+      ..value =
+          const _DurationMillisecondConverter().fromJson(json['value'] as int)
+      ..nullableValue = _$JsonConverterFromJson<int, Duration?>(
+          json['nullableValue'],
+          const _DurationMillisecondNullConverter().fromJson);
+
+Map<String, dynamic> _$JsonConverterIssue1339ToJson(
+        JsonConverterIssue1339 instance) =>
+    <String, dynamic>{
+      'value': const _DurationMillisecondConverter().toJson(instance.value),
+      'nullableValue': const _DurationMillisecondNullConverter()
+          .toJson(instance.nullableValue),
+    };
+
+Value? _$JsonConverterFromJson<Json, Value>(
+  Object? json,
+  Value? Function(Json json) fromJson,
+) =>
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: json_serializable/lib/src/type_helpers/json_converter_helper.dart [hunk 1]
```diff
@@ -202,6 +202,14 @@ _JsonConvertData? _typeConverterFrom(
     return null;
   }
 
+  if (targetType.isNullableType && matchingAnnotations.length == 2) {
```
kevmoo, line 205: What if both annotations match? What if there are more than two?

I see how this would be nice for the case you're hitting – but I worry that it has too many edge cases.

Would it be better to find the set of "better" converters in the set – and see if that set of "better" converters has a length of 1?

julek-kal: I think that is behaving better than currently and don't break other things.
It's exactly trying to find a "better" converter by its field type. When there are more than two converters or both matches it would throw "Found more than one matching converter for `$targetTypeCode`."

It's fixing 1339, so I'm able to have a JsonSerializable with converters and use it on a class that contains both nullable and non-nullable of the same type.

What other "algorithm" do you suggest to find better converters? What is the factor of "betterness" of the converter? Maybe you want to suggest to check also a `jsonType`. Ok, maybe it's a good idea, but I think that the level of complexity will be increased noticeably

t-beckmann: Maybe implement it a
```suggestion
  if (targetType.isNullableType && matchingAnnotations.length > 1) {
  matchingAnnotations
        .removeWhere((element) => !element.fieldType.isNullableType);
```
so that it is not restricted to exactly 2 matches.

Afterwards you could favor an exact type match if the result is not unique yet. Finally, instead of throwing one could the pick the first match, assuming the results are in order of declaration.

julek-kal: @t-beckmann thanks for your suggestion. @kevmoo what do you think about above solution?

jiechic: any update for this PR？

ljmatan: For whatever reason, the current workaround is to define converters both as annotations and as list members, however, it must be done separately for nullable and non-nullable types, like so: 

```dart
@JsonSerializable(
  converters: [
    ExampleSerializationHandlerString(),
    ExampleSerializationHandlerBool(),
    ExampleSerializationHandlerInt(),
    ExampleSerializationHandlerDouble(),
    ExampleSerializationHandlerMap(),
  ],
)
@ExampleSerializationHandlerStringNullable()
@ExampleSerializationHandlerBoolNullable()
@ExampleSerializationHandlerIntNullable()
@ExampleSerializationHandlerDoubleNullable()
@ExampleSerializationHandlerMapNullable()
```

This is with the latest json_serializable (6.8.0) and json_annotation (4.9.0) versions at the time of this writing.

PR #1433, opened 2024-06-27
Move to use (experimental) workspaces

kevmoo:


json_serializable/tool/readme_builder.dart
```diff
@@ -153,14 +156,21 @@ String _classCleanAndSort(Iterable<String> classes) {
 extension on BuildStep {
   AssetId assetIdForInputPackage(String path) => AssetId(inputId.package, path);
 
+  // WARNING! This function does BAD THINGS
```
kevmoo, line 159: Not my best work...but 🤷

jakemac53: Looking at this code, there might actually be a proper way to do it. The JSON annotation package _could_ make its own pubspec.yaml file public using the `additional_public_assets` option.

I think you could just add into its `build.yaml` file:

```yaml
additional_public_assets:
  - pubspec.yaml
```

And then you could read it just like `buildStep.readAsString(AssetId('json_annotation', 'pubspec.yaml'))`?

You would have to require the latest json_annotation, which adds the public asset.

jakemac53 (approved):

PR #1444, opened 2024-08-14
Fix bool? casted to bool.

qwe321:
Question mark was missing in case of optional bool field.


kevmoo:
We'll really want test coverage here!

kevmoo:
Need the CLA signed and please add a test!

qwe321:
I'm sorry I cannot reproduce it anymore on recent versions. It seems it cause was somewhere else

PR #1445, opened 2024-09-10
Bump deps on analyzer and dart_style

kevmoo:


json_serializable/tool/shared.dart
```diff
@@ -8,7 +8,8 @@ import 'package:build/build.dart';
 import 'package:dart_style/dart_style.dart';
 import 'package:yaml/yaml.dart';
 
-final formatter = DartFormatter();
+final formatter =
+    DartFormatter(languageVersion: DartFormatter.latestLanguageVersion);
```
kevmoo, line 12: Just realized this is pointless, since `latestLanguageVersion` is the default. 🤷 🤦

jakemac53: You also really don't want to do it this way. I am prepping a change to source_gen which is going to be breaking, and will give you a language version to use in the format function.

Issue #1380, opened 2023-12-07
Allow more than one value map to the same enum value with `@JsonValue`


mateusfccp:
Currently, we can use `@JsonValue` in an enum to define what value maps to it.

```dart
enum CustomType {
  @JsonValue('a')
  valueA,
  @JsonValue('b')
  valueB,
  @JsonValue('c')
  valueC,
}
```

Now, I want to map more than one value to the same enum value. I cannot do this unless I write a custom `JsonConverter` for the enum.

This issue proposes that we have a way to tell `json_serializable` that we want more than one value to be mapped to our enum value.

For instance:
```dart
enum CustomType {
  @JsonValue('valueA')
  @JsonValue('a')
  valueA,
  @JsonValue('valueB')
  @JsonValue('b')
  valueB,
  @JsonValue('valueC')
  @JsonValue('c')
  valueC,
}
```

Xazin:
I would assume a new decorator eg. `@MultiJsonValues(['valueA', 'a'])` to be more proper, or an argument on the existing decorator like so `@JsonValue('a', aliases: ['valueA', 'A'])`.

I wouldn't know if this is possible since I have not looked deeply at the code.

mateusfccp:
Maybe the "aliases" option would be better, as we should have a "canonical" reference to use in `toJson`. While `fromJson` could map from many different values to a single value, when converting this value to JSON, we can only have a single option, which would be the non-alias one.

alvarogabrielgomez:
Some updates to this?

JLabeda:
Hi, any update on this issue? 🙏

PR #1459, opened 2024-11-26
Add `aliases` to `JsonValue` to enum value to be decoded from different JSON values

mateusfccp:
Solves #1380.

Adds a parameter to `JsonValue` called `aliases`, which accepts a list of values that can be alternatively used to decode a JSON.

Example:
```dart
enum StatusCode {
  @JsonValue(200, aliases: [201, 202])
  success,
  @JsonValue(301)
  movedPermanently,
  @JsonValue(302)
  found,
  @JsonValue(500)
  internalServerError,
}
```

With the code above, all of `200`, `201` and `202` would be decoded as `StatusCode.success`.

When calling `toJson`, however, the `value` parameter is used. In this case, `200`.


json_annotation/lib/src/json_value.dart
```diff
@@ -9,5 +9,10 @@ class JsonValue {
   /// Can be a [String] or an [int].
   final dynamic value;
 
-  const JsonValue(this.value);
+  /// Optional values that can be used when deserializing.
+  ///
+  /// The elements of [aliases] must be either [String] or [int].
+  final List<Object?> aliases;
```
kevmoo, line 15: This should probably be a `Set` – since we don't care about order and we shouldn't allow duplicates.

mateusfccp: Done.

mateusfccp: Done.

kevmoo:
Ideally we'd have a couple of tests to validate the failure cases, but I get that those can be tricky to add.

mateusfccp:
@kevmoo What would be the failure cases for this PR? Can you give an example?

kevmoo:
> @kevmoo What would be the failure cases for this PR? Can you give an example?

Unsupported types in the set is the big one.
`null` isn't supported, right?
Those types of things.

mateusfccp:
> > @kevmoo What would be the failure cases for this PR? Can you give an example?
> 
> Unsupported types in the set is the big one. `null` isn't supported, right? Those types of things.

Ah, yes, the idea is to mimic `value`, so until we have proper union types in Dart, we use `Object?` but should only accept `String` and `int`.

kevmoo:
You'll need to rebase this fix.

mateusfccp:
> You'll need to rebase this fix.

I rebased to master and included the requested test.

In my local machine, a test was failing, but I didn't change anything related to it (the `*ToJson` generator).

image

mateusfccp:
Mmm, I had made a regular merge, but GitHub is saying to me that "this branch must not contain merge commits", so I rewrote the commits using a rebase instead, but it is still giving me this message.

mateusfccp:
Well, the test that did not pass locally did pass here, so we are good with it.

mateusfccp:
@kevmoo Is there something preventing the code review to proceed?

json_serializable/lib/src/enum_utils.dart
```diff
@@ -31,21 +34,45 @@ bool? enumFieldWithNullInEncodeMap(DartType targetType) {
   return enumMap.values.contains(null);
 }
 
-String? enumValueMapFromType(
+String? enumMapsFromType(
```
kevmoo, line 37: Undo this rename. Makes reviewing more tricky.

json_serializable/test/default_value/default_value.g.dart
```diff
@@ -92,9 +94,18 @@ Map<String, dynamic> _$DefaultValueToJson(DefaultValue instance) =>
...
+// ignore: unused_element
 const _$GreekEnumMap = {
   Greek.alpha: 'alpha',
   Greek.beta: 'beta',
   Greek.gamma: 'gamma',
   Greek.delta: 'delta',
 };
+
+// ignore: unused_element
+const _$GreekEnumDecodeMap = {
```
kevmoo, line 106: So we're changing the generated code and creating a NEW map, even if the dev doesn't use the new feature?

kevmoo (commented):
I don't like the repeated map EVERYWHERE for one-off feature.

Could you noodle on that a bit?

kevmoo:
FYI: I've moved this package to Dart 3.8. You'll need to rebase...

mateusfccp:
@kevmoo I've rebased the code to the `master` branch.

Let me know what you think about the discussion on `default_value.g.dart` so we can progress.

PR #1472, opened 2025-01-07
Convert all DateTimes to UTC during serialization

### LINKED ISSUE #1371: Proposal: Always convert DateTime to UTC during serialization
Since dates should pretty much always be stored in UTC would it be reasonable to automatically convert DateTime objects to UTC during serialization? Could this at least be an option in the generator? I know I could write my own DateTime converter but then I have to remember to use it every time I create a new DateTime field.

I see https://github.com/google/json_serializable.dart/pull/811, but that was abandoned. I think making this a generator setting, maybe off by default to prevent issues with existing codebases, would be ideal.

Rexios80:
Resolves https://github.com/google/json_serializable.dart/issues/1371

### DIFF
--- a/example/lib/example.g.dart
+++ b/example/lib/example.g.dart
@@ -23,8 +23,8 @@ Map<String, dynamic> _$PersonToJson(Person instance) => <String, dynamic>{
       'firstName': instance.firstName,
       if (instance.middleName case final value?) 'middleName': value,
       'lastName': instance.lastName,
-      'date-of-birth': instance.dateOfBirth.toIso8601String(),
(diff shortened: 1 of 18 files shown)


kevmoo:
I'm going to close this out. I'm not sure of a clean way to do this without a big breaking change. We'll have to chat about type support configuration or something...

Siyler1 (approved):

Siyler1 (approved):

Siyler1 (approved):
D

Siyler1 (approved):

PR #1475, opened 2025-01-31
Update README.md

travis-mortimer:
This seems to be a common question that is quite difficult to answer. The most visible solution I found previously was to use the `remove_from_coverage` package to remove generated files from the coverage report. That solution complicates CLI commands and may require updating CI/CD pipelines.

### DIFF
--- a/json_serializable/README.md
+++ b/json_serializable/README.md
@@ -287,6 +287,18 @@ targets:
           include_if_null: true
 ```
 
+To exclude generated files from coverage, you can further configure `build.yaml`.
+
+```yaml
+targets:
+  $default:
+    builders:
+      source_gen:combining_builder:
+        options:
+          preamble: |
+            // coverage:ignore-file
+```
+
 [example]: https://github.com/google/json_serializable.dart/tree/master/example
 [dart build system]: https://github.com/dart-lang/build
 [package:json_annotation]: https://pub.dev/packages/json_annotation
--- a/json_serializable/tool/readme/readme_template.md
+++ b/json_serializable/tool/readme/readme_template.md
@@ -152,6 +152,18 @@ targets:
           include_if_null: true
 ```
 
+To exclude generated files from coverage, you can further configure `build.yaml`.
+
+```yaml
+targets:
+  $default:
+    builders:
+      source_gen:combining_builder:
+        options:
+          preamble: |
+            // coverage:ignore-file
+```
+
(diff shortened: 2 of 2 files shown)


kevmoo:
@travis-mortimer – you need to update the template file and re-run build. The readme is generated!

kevmoo:
@travis-mortimer – you'll need to re-run the build, too

`dart run build_runner build`

travis-mortimer:
> @travis-mortimer – you'll need to re-run the build, too
> 
> `dart run build_runner build`

I did run that command and the generated readme was (apparently) identical to what I had previously committed. When I attempted to stage the file, it disappeared.

I also saw this error
```
[SEVERE] json_serializable:_type_test_builder on test/supported_types/type_test.dart:

RangeError (end): Invalid value: Not in inclusive range 425..2020: 29
dart:core                           _StringBase.substring
tool\test_type_data.dart 198:40     TestTypeData.testContent
tool\test_type_builder.dart 170:21  _TypeTestBuilder.build
```
and ~47 `input.type_*` files under `json_serializable\test\supported_types` were modified. None of the changes looked valid.

 Here is a sample:
image

Can you help me better understand what changes are expected from running the build after updating the template?

kevmoo:
Are you on a windows machine, @travis-mortimer ?

The readme is generated from the template. And I check that the root readme is in sync w/ the template as part of CI. So both files should be updated w/ your change. Ideally the root readme is updated by running the builder on your dev machine, but hand editing both should work, too.

Have you run `pub upgrade` first?

travis-mortimer:
> Are you on a windows machine, @travis-mortimer ?
> 
> The readme is generated from the template. And I check that the root readme is in sync w/ the template as part of CI. So both files should be updated w/ your change. Ideally the root readme is updated by running the builder on your dev machine, but hand editing both should work, too.
> 
> Have you run `pub upgrade` first?

Yes I am, though I have access to a mac if we expect running it on that machine will yield better results. I did not run `pub upgrade` first. I'm interested in getting it to work the "right" way, but you let me know if at any point you'd rather try the hand edit and I'll make the change.

travis-mortimer:
@kevmoo Are there any other changes you'd like me to make?

kevmoo:
thanks for your patience here @travis-mortimer . Trying to kick it a bit from my side...

kevmoo (approved):

PR #1478, opened 2025-02-12
Fix extra line with Dart 3.7 syntax

kevmoo:
Since this package is using 3.6, you can't see the result here
Verified with a one-off in this package and another package to make
sure nothing breaks

### DIFF
--- a/json_serializable/lib/src/decode_helper.dart
+++ b/json_serializable/lib/src/decode_helper.dart
@@ -317,27 +317,18 @@ _ConstructorData _writeConstructorInvocation(
       '$className'
       '${genericClassArguments(classElement, false)}'
       '$constructorExtra(',
-    );
-  if (constructorArguments.isNotEmpty) {
-    buffer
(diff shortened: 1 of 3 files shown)


json_serializable/lib/src/decode_helper.dart
```diff
@@ -317,27 +317,18 @@ _ConstructorData _writeConstructorInvocation(
...
-      ..writeAll(namedConstructorArguments.map((paramElement) {
-        final value =
-            deserializeForField(paramElement.name, ctorParam: paramElement);
-        return '      ${paramElement.name}: $value,\n';
-      }));
-  }
-
-  buffer.write(')');
+    )
+    ..writeAll(constructorArguments.map((paramElement) {
```
kevmoo, line 321: The only real change here is the removal of the extra `..writeln()` but figured I'd do more general cleanup, too!

jakemac53 (approved):

Issue #1342, opened 2023-08-04
[Feature request] Support dart3 sealed class


lifeapps-42:
Thank you for this package!

I failed to find any discussions about this and it feels like essential feture at this moment.

Do you have any plan to implement union-like sealed class serialization, so we can avoid using freezed?

We could use this feature something like this (inspired by freezed):

```dart
@JsonSerializable(unionKey: 'subtype') // dafualt 'runtimeType'
sealed class RootSealedClass {
  const RootSealedClass();

  factory RootSealedClass.fromJson(Map<String, dynamic> json) =>
      _$RootSealedClassFromJson(json);

  @JsonValue('first') // or literal class name as default
  const factory RootSealedClass.first(String someAtribute) =
      FirstSealedClassSubtype;

  @JsonValue('second') // or literal class name as default
  const factory RootSealedClass.second(int someOtherAtribute) =
      SecondSealedClassSubtype;

  Map<String, dynamic> toJson();
}

@JsonSerializable()
class FirstSealedClassSubtype extends RootSealedClass {
  const FirstSealedClassSubtype(this.someAtribute);

  factory FirstSealedClassSubtype.fromJson(Map<String, dynamic> json) =>
      _$FirstSealedClassSubtypeFromJson(json);

  final String someAtribute;

  @override
  Map<String, dynamic> toJson() => _$FirstSealedClassSubtypeToJson(this);
}

@JsonSerializable()
class SecondSealedClassSubtype extends RootSealedClass {
  const SecondSealedClassSubtype(this.someOtherAtribute);

  factory SecondSealedClassSubtype.fromJson(Map<String, dynamic> json) =>
      _$SecondSealedClassSubtypeFromJson(json);

  final int someOtherAtribute;

  @override
  Map<String, dynamic> toJson() => _$SecondSealedClassSubtypeToJson(this);
}
```

kevmoo:
Maybe? How does Freezed do this?

lifeapps-42:
FromJson methods for subtypes are generated as usual:

```dart
FirstSealedClassSubtype _$FirstSealedClassSubtypeFromJson(Map<String, dynamic> json) =>
    FirstSealedClassSubtype(
      someAtribute: json['someAtribute'] as String,
    );

SecondSealedClassSubtype _$SecondSealedClassSubtypeFromJson(Map<String, dynamic> json) =>
    SecondSealedClassSubtype(
      someOtherAtribute: json['someOtherAtribute'] as int,
    );
[...]

PR #1483, opened 2025-03-30
feat: implement first version of sealed class serialization

O-Hannonen:
## Whats included

This PR is a proposal related to #1342. Since the issue asked for a PR only to evaluate if adding this functionality is worth the added complexity, this PR DOES NOT (yet) include:

- tests for the new functionality
- documentation for the new functionality
- and consideration for all the different edge cases (eg. error handling for unsupported config combos)

I'd be happy to contribute towards those too if the maintainers give green light. 

## Things to discuss / consider

- Should `genericArgumentFactories` be supported and how? Or should it warn / throw as invalid config combo?
- Should there be a `UnionKey` annotation similar to `JsonKey` allowing naming change per sealed implementation?
- Should the generator warn / throw if discriminator is also present as sealed class implementations field? Would lead to conflict in generated `toJson` method where class fields are added to the output with spread operator


json_serializable/lib/src/decode_helper.dart
O-Hannonen, line 1: Was having a bit hard time understanding all the different things happening in `createFactory`. As the sealed class deserialization would have made it even more complex, decided to split it up into smaller functions for enhanced readability.

kevmoo: usually a good idea 😄

json_serializable/lib/src/decode_helper.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:core';
+
```
kevmoo, line 6:
```suggestion
```

Shouldn't be needed.

json_serializable/lib/src/decode_helper.dart
```diff
@@ -27,12 +29,94 @@ mixin DecodeHelper implements HelperCore {
...
+    );
+  }
+
+  /// Creates the function signature around [functionBodyParts]
+  /// that will be used to deserialize the class.
+  ///
+  /// If [functionBodyParts] has only one element, expression body is used.
+  ///
+  /// ```dart
+  /// '''
```
kevmoo, line 96: shouldn't have `'''` here. Just the dart code should be fine.

json_serializable/lib/src/decode_helper.dart
```diff
@@ -49,114 +133,204 @@ mixin DecodeHelper implements HelperCore {
...
-          .map((fe) => fe.name)
-          .toList(),
-      unavailableReasons,
-      deserializeFun,
-    );
+  /// Creates the body of the function that deserializes a sealed class.
+  ///
+  /// For example:
+  /// ```dart
+  /// '''
```
kevmoo, line 153: ditto

json_serializable/lib/src/utils.dart
```diff
@@ -162,6 +168,24 @@ ConstructorElement constructorByName(ClassElement classElement, String name) {
...
+          const TypeChecker.fromRuntime(JsonSerializable),
+        )
+        .map((e) => e.element)
+        .whereType<ClassElement>()
+        .where(
+          (e) => e.allSupertypes.contains(sc.thisType),
+        );
+  }
+
+  return [];
```
kevmoo, line 186: If you want to be fancy 😄 

```suggestion
  return const Iterable<ClassElement>.empty();
```

kevmoo (commented):
I'm impressed, yo. A lot of folks have asked for this.

I think I like your approach. You're absolutely right, the devil is in the details here...lots of overlapping issues.

I'd avoid the generic factories for now. It'll likely be a HUGE mess that won't affect most folks.

If you look at many of my previous PRs, there is usually more code in tests than there is in the actual implementation.

If you're game to run with what you have, I'm happy to support you. Very cool!

json_annotation/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: json_annotation
-version: 4.9.1-wip
+version: 4.10.0
```
O-Hannonen, line 2: `4.10.0-wip`?

json_serializable/README.md
```diff
@@ -310,15 +359,15 @@ targets:
...
-[`JsonConverter`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonConverter-class.html
-[`JsonEnum.valueField`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonEnum/valueField.html
-[`JsonEnum`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonEnum-class.html
-[`JsonKey.fromJson`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/fromJson.html
-[`JsonKey.toJson`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/toJson.html
-[`JsonKey`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey-class.html
-[`JsonLiteral`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonLiteral-class.html
-[`JsonSerializable`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonSerializable-class.html
-[`JsonValue`]: https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonValue-class.html
+[`JsonConverter`]: https://pub.dev/documentation/json_annotation/4.10.0/json_annotation/JsonConverter-class.html
```
O-Hannonen, line 362: Did I do something wrong when these show as changes here?

json_serializable/lib/src/generator_helper.dart
```diff
@@ -51,6 +52,80 @@ class GeneratorHelper extends HelperCore with EncodeHelper, DecodeHelper {
...
+          throw InvalidGenerationSourceError(
+            'The class `${element.displayName}` is sealed but its '
+            'subclass `${sub.displayName}` is not annotated with '
+            '`JsonSerializable`.',
+            todo: 'Add `@JsonSerializable` annotation to ${sub.displayName}.',
+            element: sub,
+          );
+        }
+      });
+    }
```
O-Hannonen, line 127: All of the other checks we can do when generating subclass but not this one as the generator will never get there for the subclass if its not annotated. IMO would be cleaner if all the checks would be done with lookups from subclass to superclass (would not need to use the `sealedSubClasses` method here yet). 

Should `sealedSubClasses` be passed to decode/encode helpers if its already evaluated here? Now its called again in the helpers...

json_serializable/test/json_serializable_test.dart
```diff
@@ -122,11 +122,72 @@ const _expectedAnnotatedTests = {
...
   'SubTypeWithAnnotatedFieldOverrideExtends',
   'SubTypeWithAnnotatedFieldOverrideExtendsWithOverrides',
   'SubTypeWithAnnotatedFieldOverrideImplements',
+  'SubUnionRenameKebab',
+  'SubUnionRenameNone',
+  'SubUnionRenamePascal',
+  'SubUnionRenameScreamingSnake',
+  'SubUnionRenameSnake',
+  'SubWithConflictingDiscriminatorWithDefaultNameExt',
+  'SubWithConflictingDiscriminatorWithDefaultNameImpl',
```
O-Hannonen, line 152: There is some overlap in these as they test both the use of `extends` and `implements` keywords. The analyzer treats them differently within `ClassElement` (one comes from `.interfaces` and one from `.supertype`) so there is a risk that some refactor would break one but not other

O-Hannonen:
@kevmoo Did I miss any type of tests that would need to be added? Or combos of unsupported configurations?

json_annotation/pubspec.yaml
kevmoo: yep!

json_serializable/README.md
kevmoo: If you change the release to be something other than `-wip` it switches to the version number instead of `latest` So fixing adding `-wip` will fix.

json_serializable/test/json_serializable_test.dart
kevmoo: glad to have both! good thinking!

kevmoo:
you'll need to rebase here. I'd wait until my Dart 3.8 change lands, though.

json_annotation/lib/src/json_serializable.dart
```diff
@@ -31,6 +31,26 @@ enum FieldRename {
   screamingSnake,
 }
 
+/// Values for the automatic class renaming behavior for [JsonSerializable]
+/// with sealed classes.
+enum UnionRename {
```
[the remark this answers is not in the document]
O-Hannonen: My thought process was that it might be confusing for the package users if union rename is of type `FieldRename`, since its renaming class names and not any field names. Although after the renaming the string is used as key of a map so there the field would make a bit more sense again.

Technically there shouldn't be a limitation for not just using `FieldRename` since they have the same rename options (at least now, probably no need to ever have different options). So should we remove the `UnionRename`?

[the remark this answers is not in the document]
kevmoo: Our users will be fine. A *better* thing would probably be to rename `FieldRename` to just be `RenameType` and then `typedef FieldRename = RenameType;` and mark as deprecated.

Just put a `todo` in for that and I can follow-up.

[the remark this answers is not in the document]
O-Hannonen: Sounds good, this is now done. Super nice and handy the typedef deprecation trick!

json_annotation/lib/src/json_serializable.dart
```diff
@@ -12,8 +12,12 @@ import 'json_key.dart';
 
 part 'json_serializable.g.dart';
 
+// TODO: Remove typedef
```
kevmoo, line 15: This will have to be documented in the changelog, etc!

kevmoo:
willing to do a rebase here?

hai-trung-le:
Any update?

kevmoo (changes requested):
The amount of code churn here is HUGE!

(1) Did you do other, unrelated refactoring? Please minimize this so it's easier to review
(2) Could you write up a 1-page review of what you're proposing
  - What does this enable
  - What does a user of json_serializable have to do to make this work
  - What is changed in how json_serial is implemented to make it happen

I'm not against these changes, but I need some front-loading!

O-Hannonen:
The large diff is primarily due to extensive integration tests and examples for the new sealed class support. I did refactor `decode_helper.dart` and `encoder_helper.dart` to ensure the code is readable even after this "bigger" new feature. I can totally revert the refactoring if that's preferred, but I think the sealed class implementation would become a bit messy then.

## Sealed Class Support Overview

What it enables:
Native support for Dart 3 sealed classes. json_serializable can now generate a central fromJson factory on a base sealed class that dispatches to subclasses using a discriminator field (defaulting to type), and a toJson that includes this discriminator.

User Requirements:
1. Annotate the base sealed class and all subclasses with @JsonSerializable().
2. Add fromJson and toJson to the base class that call the generated _$BaseClassFromJson and _$BaseClassToJson.
3. Subclasses maintain their own standard fromJson/toJson factories.
4. Customization is available via unionDiscriminator and unionRename options.

Implementation Changes:
1. Annotation Update: Added unionDiscriminator and unionRename to JsonSerializable.
2. Validation: GeneratorHelper now validates sealed hierarchies for consistent annotations and configurations.
3. Refactored Logic: DecodeHelper and EncodeHelper generate switch statements for polymorphic dispatching of sealed classes
4. Analyzer Utilities: Added helpers in utils.dart to navigate sealed structures using the analyzer API.

PR #1519, opened 2025-08-20
Add `dart format off` and `dart format on` comments

### LINKED ISSUE #1518: The `page_width` option in the analysis_options.yaml is ignored
json_serializable disregards the `page_width` setting in `analysis_options.yaml` and uses the default value of 80. Therefore, if a different `page_width` is specified in the project's `analysis_options.yaml`, a separate formatting step is required after running `build_runner build`.

https://github.com/google/json_serializable.dart/blob/83e79058f3da3f912965dd9356796f25c41ae1b3/json_serializable/lib/src/json_part_builder.dart#L113-L114

### Steps to Reproduce

This behavior can be reproduced in the `json_serializable.dart` project.

1. Set a `page_width` value other than 80 in `/analysis_options.yaml`:
   ```yaml
   formatter:
     page_width: 120
   ```
2. Navigate to the `/example`.
3. Run `dart run build_runner build -d`.
   * The generated code is formatted with page_width `80`.
4. Run `dart format lib/`.
   * The generated code is formatted with page_width `120`.
5. Run `dart run build_runner build -d`.
   * The generated code is formatted with page_width `80`.

### Proposal

How about adding the line `// dart format width=80` to the generated code?

https://github.com/dart-lang/dart_style/blob/bacc9bf8fcec6fffc15ad7e9257b7faba5298ea9/CHANGELOG.md#300

### Trailing comma

As reported in https://github.com/google/json_serializable.dart/issues/1486, the trailing commas option is also not reflected. However, since commas are removed when formatted with `TrailingCommas.automate`, I cannot think of any cases where additional formatting would be necessary.

koji-1009:
close #1518 

Added the comments `// dart format off` and `// dart format on` to the code generated using `defaultFormatOutput`.
With this fix, additional formatting will no longer be applied to the generated code even if the project's `page_width` setting is not 80.

ref https://github.com/dart-lang/build/issues/3812#issuecomment-2641576908
ref https://github.com/dart-lang/build/issues/3887#issuecomment-2687338332

### DIFF
--- a/example/lib/example.g.dart
+++ b/example/lib/example.g.dart
@@ -2,6 +2,8 @@
 
 part of 'example.dart';
 
+// dart format off
+
 // **************************************************************************
 // JsonSerializableGenerator
 // **************************************************************************
@@ -86,3 +88,5 @@ final _$glossaryDataJsonLiteral = {
     },
   },
 };
+
+// dart format on
--- a/example/lib/generic_response_class_example.g.dart
+++ b/example/lib/generic_response_class_example.g.dart
@@ -2,6 +2,8 @@
 
 part of 'generic_response_class_example.dart';
 
+// dart format off
+
 // **************************************************************************
 // JsonSerializableGenerator
 // **************************************************************************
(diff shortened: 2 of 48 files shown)


kevmoo:
Interesting idea, but I think this should be a feature added at either `pkg:build` and friends or at least `pkg:source_gen`

koji-1009:
Thanks for the review. OK, I'll wait for this feature to be added to `pkg:build` and friends.

PR #1531, opened 2025-10-21
Allow an option called `run_only_if_triggered`.

davidmorgan:
Thinking about using the new(ish) "triggers" feature in json_serializable, it actually turns out it doesn't work yet because json_serializable is strict about its options :)

https://github.com/dart-lang/build/discussions/4254#discussioncomment-14739442

This allows the option, but does not actually use it (yet). People can then enable it in their own `build.yaml`.

### DIFF
--- a/json_serializable/test/config_test.dart
+++ b/json_serializable/test/config_test.dart
@@ -37,7 +37,8 @@ void main() {
   });
 
   test('triggers config is ignored', () async {
-    // This key is used by build_runner so it's meaningful for any builder.
+    // This key is used by build_runner so it's meaningful for any builder,
+    // check it doesn't cause json_serializable's config validation to throw.
     jsonSerializable(const BuilderOptions({'run_only_if_triggered': true}));
   });
 


json_serializable/test/config_test.dart
```diff
@@ -36,6 +36,11 @@ void main() {
     expect(builder, isNotNull);
   });
 
+  test('triggers config is ignored', () async {
+    // This key is used by build_runner so it's meaningful for any builder.
+    jsonSerializable(const BuilderOptions({'run_only_if_triggered': true}));
```
kevmoo: Uh...is any result measured or anything?

davidmorgan:
I think it's good to go like this, it's already useful even if released without actually using triggers, because it lets users turn them on. Thanks!

json_serializable/test/config_test.dart
davidmorgan: Nope ... before this PR it throws, so doing nothing is an improvement ;)

I can call that out more explicitly if you like? :)

PR #1532, opened 2025-11-14
Skip warning about duplicate annotations if the values match.

davidmorgan:
Fix https://github.com/rrousselGit/freezed/issues/1289

Freezed copies annotations onto generated code, which triggers the json_serializable warning.

Fix by skipping warning if the value matches.

I'm not 100% sure about the implementation, I did add a test, but please do check it makes sense :)


kevmoo (approved):
Thanks! Want me to publish soonish?

PR #1552, opened 2026-01-31
generic support

ManuelLeitner:
add support for generic JsonConverter, toJson and fromJson with arbitrary generic arguments


json_serializable/lib/src/type_helper_ctx.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/constant/value.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/dart/constant/value.dart';
```
kevmoo: This is NOT ideal...

kevmoo (commented):
at a minimum we need a changelog entry and tests here!

PR #1561, opened 2026-02-18
Improve error message when generating code for Record types with unsupported fields

### LINKED ISSUE #1559: `typeToCode` breaks when presented with records
Image

This is happening where some of my models have records as member types.

I would suggest creating a TypeVisitor in order to correctly represent all accepted types in code without erroring.
At the very least json_serializable can give a useful error.

`json_serializable: ^6.12.0`

kevmoo:
Fixes #1559

json_serializable/lib/src/utils.dart: Updated typeToCode to properly handle RecordType. Now correctly formats record types (including nullable and nested records) and falls back to getDisplayString() for unsupported types like FunctionType instead of throwing UnimplementedError.

json_serializable/test/src/_json_serializable_test_input.dart: Added regression tests for records with unsupported types and multiple converters on records.

json_serializable/test/json_serializable_test.dart: Added new tests to the expected test list.

json_serializable/CHANGELOG.md: Added changelog entry.


json_serializable/test/src/_json_serializable_test_input.dart
```diff
@@ -626,3 +626,138 @@ class DateTimeUtcTestClass {
 
   DateTimeUtcTestClass(this.date);
 }
+
+@ShouldThrow(
+  'Could not generate `fromJson` code for `record` because of type '
```
kevmoo: Too many test cases in here!

Really only need one for the bad type and one for the converter bit!

natebosch (approved):

PR #1571, opened 2026-04-30
Update analyzer to latest, prepare v6.13.2

kevmoo:


json_serializable/lib/src/field_helpers.dart
```diff
@@ -86,8 +86,7 @@ List<FieldElement> createSortedFieldSet(ClassElement element) {
       continue;
     }
 
-    if (v is GetterElement) {
-      final variable = v.variable as FieldElement;
+    if (v case GetterElement(variable: final FieldElement variable)) {
```
kevmoo, line 89: I do worry about the readability of this for beginners...

natebosch: These patterns have been much harder to write than to read - I would not expect reading this syntax to trip anyone up.

natebosch (approved):

PR #1587, opened 2026-08-12
Use $enumEncode for non-nullable enum toJson lookups

### LINKED ISSUE #1515: Unnecessary use of a null check ('!') on enum values.
Latest version (6.10.0) generates the following code for an enum field

```dart
Map<String, dynamic> _$XyzToJson(Xyz instance) =>
    <String, dynamic>{
      'state': _$AbcEnumMap[instance.state]!,
    };
```

The generated `_$AbcEnumMap` is a const map with all the enums listed, and the linter now complains that it is an unnecessary operation.

AzazelSensei:
## Summary
Generated `toJson` for non-nullable enums currently emits `_$EnumMap[instance.field]!`. When that expression is stored in a `Map<String, dynamic>`, analyzers that enable `unnecessary_null_checks` flag the bang as unnecessary.

This adds a small `$enumEncode` helper (alongside `$enumDecode`) and uses it for non-nullable enum encoding so generated call sites stay lint-clean while still producing a non-nullable value for Map keys and other typed contexts (#1145).

Fixes #1515

## Test plan
- [x] `dart test --exclude-tags=presubmit-only` in `json_serializable` (660 passed)
- [x] Integration regression for #1515 + existing #1145 map-key typing test
- [x] Regenerated committed `.g.dart` outputs


AzazelSensei:
Closing this after 15 days without a comment.

